PR #13, opened 2019-12-12
Add BigWarpLauncher

tischi:
Hi @NicoKiaru 

If you are interested you could have a look and comment how this compares to your approach.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,30 @@
 			<artifactId>bigdataviewer-vistools</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>net.imagej</groupId>
+			<artifactId>imagej-legacy</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>net.imagej</groupId>
+			<artifactId>imagej</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>net.imagej</groupId>
+			<artifactId>imagej-common</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>net.imagej</groupId>
+			<artifactId>ij</artifactId>
+		</dependency>
+
+		<!-- BigWarp -->
+		<dependency>
+			<groupId>sc.fiji</groupId>
+			<artifactId>bigwarp_fiji</artifactId>
+			<version>5.1.1-SNAPSHOT</version>
+		</dependency>
+
 		<!-- Test dependencies -->
 		<dependency>
 			<groupId>junit</groupId>
--- a/src/main/java/sc/fiji/bdv/BdvUtils.java
+++ b/src/main/java/sc/fiji/bdv/BdvUtils.java
@@ -1,0 +1,77 @@
+package sc.fiji.bdv;
+
+import bdv.util.Bdv;
+import bdv.viewer.Source;
+import bdv.viewer.state.SourceState;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class BdvUtils
+{
+	public static int getSourceIndex( Bdv bdv, Source< ? > source )
+	{
+		final List< SourceState< ? > > sources =
+				bdv.getBdvHandle().getViewerPanel().getState().getSources();
+
+		for ( int i = 0; i < sources.size(); ++i )
+			if ( sources.get( i ).getSpimSource().equals( source ) )
+				return i;
+
+		return -1;
+	}
+
+	public static Source< ? > getSource( Bdv bdv, int sourceIndex )
+	{
+		final List< SourceState< ? > > sources =
+				bdv.getBdvHandle().getViewerPanel().getState().getSources();
+
+		return sources.get( sourceIndex ).getSpimSource();
+	}
+
+	public static Source< ? > getVolatileSource( Bdv bdv, int sourceIndex )
+	{
+		final List< SourceState< ? > > sources =
+				bdv.getBdvHandle().getViewerPanel().getState().getSources();
+
+		return sources.get( sourceIndex ).asVolatile().getSpimSource();
+	}
+
+	public static double[] getDisplayRange( Bdv bdv, int sourceId )
+	{
+		final double displayRangeMin = bdv.getBdvHandle().getSetupAssignments()
+				.getConverterSetups().get( sourceId ).getDisplayRangeMin();
+		final double displayRangeMax = bdv.getBdvHandle().getSetupAssignments()
+				.getConverterSetups().get( sourceId ).getDisplayRangeMax();
+
+		return new double[]{ displayRangeMin, displayRangeMax };
+	}
+
+	public static String getSourceName( Bdv bdv, int sourceId )
+	{
+		return bdv.getBdvHandle().getViewerPanel().getState().getSources()
+				.get( sourceId ).getSpimSource().getName();
+	}
+
+	public static int getSourceIndex( Bdv bdv, String sourceName )
+	{
+		return getSourceNames( bdv ).indexOf( sourceName );
+	}
+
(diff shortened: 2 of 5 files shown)


NicoKiaru (commented):
Hi @tischi, 

I did not have time to fully review it, but here are my thoughts:
* Limitation: Only one moving source and one fixed source
* Limitation: Why only an affine transform ? Why not letting the choice to the user ?
* How would you envision the selection of multiple sources for fixed and moving sources ? And that could be potentially in two separate bdv handles ?
* Advice : if you transfer `SourceAndConverter` instead of `SpimSource`, there is no need to handle the volatile/non volatile view.
* Q : why do you need to add a listener ? Sorry I did not understand everything in the commit because e I'm lacking a bit of time right now

tischi:
Thanks for the suggestions!
Do you think it makes sense that I implement your suggestions into my code, which I am happy to do, or is it more efficient for us to use your code here?

NicoKiaru:
Just copy whatever you need. No issue, I'd be pleased to try it myself but I'm still struggling to understand the logic behind behaviour and input/action map. I'm working on it, but for now I cannot contribute to your repo. Conversely, if you do not understand something in https://github.com/BIOP/bigdataviewer_scijava/blob/1c5c1e9fc06de0e93ccc8cb30de54a995ad2cc78/src/main/java/ch/epfl/biop/bdv/scijava/command/edit/register/BigWarpInitWithBdvSources.java , just let  me know.

But on a general manner, I think you/we need to decide what would be the way of passing an input which would be a list of  bdv Source. 
* One way to do it is to pass a `BdvHandle` and a `String` defining the index of the sources. -> good thing about this is that we know which BdvHandle they originate from. That was my choice. But it's not optimal. If the user removes a `Source` from a BdvHandle, then the indexation should change ? or not ? How to listen ?
* Another way is to pass a `List<Source>` or a `List<SourceAndConverter>`, but then : are they attached to none/one/many `BdvHandle`

Generally speaking, how would you like to deal with multiple  bdv sources and multiple `BdvHandle` ?

Because a completely different manner is just to hold a list of bdv sources somewhere, independently of any `BvdHandle`, and then keep for each Source (or SourceAndConverter) a list of `BdvHandle` where they are displayed, in order to update when necessary.

I'm sorry I couldn't come in Dresden, so maybe you have discussed this before and made some choice already. Do you ? What have they been ?

tischi:
> One way to do it is to pass a BdvHandle and a String defining the index of the sources. -> good thing about this is that we know which BdvHandle they originate from. That was my choice. But it's not optimal. If the user removes a Source from a BdvHandle, then the indexation should change ? or not ? How to listen ?

For exactly this reason, Tobias is working on getting rid of the indexing logic and rather to refer to `Source` objects directly instead. However, when this code becomes useable for us it unclear.

> Another way is to pass a List<Source> or a List<SourceAndConverter>, but then : are they attached to none/one/many BdvHandle

My preferred choice would be to pass `bdvHandle` **and** `List<SourceAndConverter>`. What would you think? 

1. How do you get the `SourceAndConverter` from a `bdvHandle`? Like this? 
```
final List< SourceState< ? > > sources = bdv.getBdvHandle().getViewerPanel().getState().getSources();
```

tischi:
Another thing:

> I'd be pleased to try it myself but I'm still struggling to understand the logic behind behaviour and input/action map.

The point here is that one does not have to understand this logic to contribute. The idea is to contribute pure functionality, such as this class: https://github.com/haesleinhuepf/bigdataviewer-playground/blob/master/src/main/java/sc/fiji/bdv/sources/read/SourceLoaderAndAdder.java#L21
How this is later put into the UI is another layer.

So, for example, if you provide a class:
```
class BigWarpSomething
{
BigWarpSomething( BdvHandle bdvHandle, List< SourceAndConverter > fixedList, List< SourceAndConverter > movingList )
}
```

This would be sufficient.

NicoKiaru:
> Like this?
```
final List< SourceState< ? > > sources = bdv.getBdvHandle().getViewerPanel().getState().getSources();
```
Exactly. And BigWarp can be initialized with `SourceAndConverter`s. Looks like you can also transfer the `ConverterSetup`.

> The point here is that one does not have to understand this logic to contribute. 

Yes but I'd like to understand in order to add (on a separate repo) the scijava layer. But ok, it's not the most important right now.

> My preferred choice would be to pass bdvHandle and List<SourceAndConverter>. What would you think?

This is probably unsufficient if your `SourceAndConverter` is displayed in several `BdvHandle`. An `Observer` pattern could be an option : it's a way to synchronize multiple display for instance. If one  of them is modified, all the windows are notified. (https://en.wikipedia.org/wiki/Observer_pattern)

I know it looks painful, but whether we want to be able to handle multiple BdvHandle is critical. We can make the choice of supporting only one (could be limiting : launching BigWarp from an existing `BdvHandle` already gives 3 windows)

tischi:
>  it's a way to synchronize multiple display for instance. 

I think we do should support this! Could you look how it is done in BigWarp?
I would go via a listener-observer pattern. `ViewerPanel` has some of those methods:

```
	/**
	 * These listeners will be notified about changes to the
	 * {@link #viewerTransform}. This is done <em>before</em> calling
	 * {@link #requestRepaint()} so listeners have the chance to interfere.
	 */
	protected final CopyOnWriteArrayList< TransformListener< AffineTransform3D > > transformListeners;

	/**
	 * These listeners will be notified about changes to the
	 * {@link #viewerTransform} that was used to render the current image. This
	 * is intended for example for {@link OverlayRenderer}s that need to exactly
	 * match the transform of their overlaid content to the transform of the
	 * image.
	 */
	protected final CopyOnWriteArrayList< TransformListener< AffineTransform3D > > lastRenderTransformListeners;

	/**
	 * These listeners will be notified about changes to the current timepoint
	 * {@link ViewerState#getCurrentTimepoint()}. This is done <em>before</em>
	 * calling {@link #requestRepaint()} so listeners have the chance to
	 * interfere.
	 */
	protected final CopyOnWriteArrayList< TimePointListener > timePointListeners;

	protected final CopyOnWriteArrayList< InterpolationModeListener > interpolationModeListeners;

```

tischi:
...maybe at some point we could also implement a `DisplaySettingsListener` such that B&C changes could be communicated (I think this does not exist yet, does it?).

NicoKiaru:
If you have the same `SourceAndConverter` in two windows, this is automatically done. Two issues nonetheless : 
1 - both windows need to repaint. This is not automatically done, but you can 'drag' a little bit both windows to update them. 
2 - If you copy only SourceAndConverter from one window to the other, the `ConverterSetup` is still attached to its original window, and not transfered to the second window. So you can set the B&C only from the first window... and if you close it, you loose it.

NicoKiaru:
I would like to try another way of launching BigWarp on a different branch:

My idea is the following : not specifying `BdvHandle` as an input, but `Sources` only.

Then in the output, you will get:

* The BdvHandle of the moving sources
* The BdvHandle of the fixed sources
* A way to get the current BigWarp transform

What's important in the associated BigWarp Command, is to put the newly created BdvHandles into the Object Service, so that you can work on it like any other Bdv Window

This should make everything accessible. Good for you ?

tischi:
> I would like to try another way of launching BigWarp on a different branch

Please do it, because my solution does not work at the moment as it would require a possibility to add a `TransformListener` to BigWarp, which is only possible in a version that we do not yet have as a maven dependency.

NicoKiaru:
I'd suggest closing this branch because it has conflicts and got some limitations mentioned before. @tischer if you agree I let you do that.
See https://github.com/haesleinhuepf/bigdataviewer-playground/pull/39 for an alternative option on BigWarp integration.

PR #24, opened 2019-12-15
Refactor BrightnessAdjuster

tischi:
I made a BrightnessAdjuster class: 
https://github.com/haesleinhuepf/bigdataviewer-playground/blob/c3c1a70bca60eee10f7a075370bde70adad6f59d/src/main/java/sc/fiji/bdv/sources/display/BrightnessAdjuster.java#L23


NicoKiaru (approved):
My first review, and I don't know what to write, but cool!

PR #25, opened 2019-12-15
Demo of using java functional interface in bdv actions

NicoKiaru:
For demo purpose, I implemented an action (apply an AffineTransform on a Source) which is also implementing a [Java Functional interface](https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html) `Function<Source,Source>`. A [Function](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html) takes an input and returns a single output. Here a `Source` as an input, and a `Source` as an output.
* this keeps the general scheme that we decided for action (implementing Runnable, etc)
```
 SourceAffineTransform affineTransformer = new SourceAffineTransform(aSource, at3d);
```
You can transform a single `aSource` as usual:
```
affineTransformer.run();
affineTransformer.getSourceOut();
```

But since this action also implements a functional interface `Function<Source, Source>`, after its initialization, you can also 'batch' its action on a List of `Source`s (`sourcesIn`):
```
List<Source> sourcesTransformed = Lists.transform(sourcesIn, affineTransformer::apply);
```

And it can be done easily in parallel if required:
```
List<Source> sourcesTransformed = sourcesI.parallelStream().map( i -> affineTransformer.apply(i)).collect(Collectors.toList());
```

This does not require extra dependency. Pure Java.

Also adding a `Source` to a `BdvHandle` is another kind of action which can benefit from implementing another functional interface : since it takes one input and has no output : it's a [`Consumer`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html).

All of these concepts are demo in the test class [`AffineTransformSourceAndConverterBatchDemo`](https://github.com/NicoKiaru/bigdataviewer-playground/blob/bdv_scijava_import/src/test/src/sc/fiji/bdv/sourceAndConverter/AffineTransformSourceAndConverterBatchDemo.java)

Try it and let me know what you think. If you want to avoid accepting this PR, it's all in the bdv_scijava_import branch here : https://github.com/NicoKiaru/bigdataviewer-playground/tree/bdv_scijava_import

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,10 @@
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>sc.fiji</groupId>
+            <artifactId>bigwarp_fiji</artifactId>
+        </dependency>
+    </dependencies>
 
 </project>
--- a/src/main/java/sc/fiji/bdv/BdvUtils.java
+++ b/src/main/java/sc/fiji/bdv/BdvUtils.java
@@ -1,8 +1,6 @@
 package sc.fiji.bdv;
 
 import bdv.VolatileSpimSource;
-import bdv.tools.brightness.MinMaxGroup;
-import bdv.tools.brightness.SetupAssignments;
 import bdv.tools.transformation.TransformedSource;
 import bdv.util.BdvHandle;
 import bdv.util.BdvStackSource;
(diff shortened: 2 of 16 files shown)


NicoKiaru:
Cool, thanks for merging!

FYI I think I may have destroyed iniBrightnessAndContrast. I don't know what happened but the method was not present in BdvUtils...

PR #28, opened 2019-12-15
Refactor source adder

tischi:
Hi @NicoKiaru,

1. What do you think in general?

https://github.com/haesleinhuepf/bigdataviewer-playground/blob/refactorSourceAdder/src/main/java/sc/fiji/bdv/ui/SingleSourceLoaderCommand.java#L23

2. Do you want to add your code for showing the `bdvHandle` `Parameter`, in case it was not resolved before? If you also put it into the `ui` package it would be easy to remove it later.... I think it would be cool. Would it help you using this function in your repo?


NicoKiaru:
Ok, I'll train to do a proper review, even if it's a dummy commit. Let's see (this will take a bit of time)

tischi:
Not that I ever did a proper review ;-)
I already see a lot of good things happening by us working on this together. I am super happy! 
Thanks for joining in!!!!

NicoKiaru:
Ok, one limitation is that a `SpimData` can have many `Source`s (I have situations with >100 sources). So the autocontrast is too limited if it's on Source 0.

But more importantly: the `Source`, when stored within a `SpimData` object, have potentially many associated metadata (all transforms, `Tile`, `Angle`, but that's extensible, they are all of class `Entity`). When you show a `SpimData` object into a `BdvHandle` using `BdvFunctions`, you basically lose these metadata, and you lose the opportunity at the same time to modify and resave your `SpimData` object after a registration for instance.

So, here's a major design decision:
* Should we open a `SpimData` and drop it all into a `BdvHandle` ? How to find to sources indexes back between `SpimData` and the `BdvHandle` ?
* Or should we just open the `SpimData` object into another window (no viewer, just a tree view for instance) ? Then we can show it all into a `BdvHandle` or just a subset of Sources ?

I'm in favor of the second option personally. This is super convenient to adjust the B&C and/ or the color of a list of related `Source`s.

But we can also write both actions ( a simple that opens and show, and a more adavanced which just display somehow a SpimData object), and see after usage what's necessary or convenient or not.

tischi:
Frankly, I avoided SpimData so far at all.

My thoughts are:
- SpimData, as the name suggests, was initially intended to hold SPIM, i.e. light-sheet data, where all this complexity makes sense. 
- Personally I like much better the concept of a `Source` which is 3D volumes over time. 

Question:
- Do you know whether a `SpimData` object could be "lossless" mapped onto a `List<SourceAndMetadata>`, where `Metadata` would hold all the extra informations?

One reason I ask is because @hanslovsky in Paintera seems to follow a philosophy of not dealing with `SpimData` at all but rather having `Source` objects that hold much more information than a plain `Source`; I think he calls it `SourceState`, I started something in this direction, calling it `SourceAndMetadata`.

So, if we could do this `SpimData` => `List<SourceAndMetadata>` mapping, could we not just show all of them in Bdv? (and adjust the B&C for all of them and also adjust the ViewerTransform such that all of them are visible (I could work on this)).

And regarding your second option: I think it is a great to show this extra window.
1. Would you like to add the code for this? 
1. Could you make it such that it is optional whether his window pops up? Because, as said, I think Tobias is working on the groupings of `Source`s and also improving the Bdv UI in that respect. Once he has done that it might be that we could use his stuff instead.

NicoKiaru:
Initially I was also rebuted by `SpimData` (and the name is obsolete, but not its function), but honestly there are some nice features about this:
* it's already existing - no need to write something on our side
* it's compatible with existing plugin (including BigStitcher)
* it can be extended to store any metadata or to use different readers (potentially n5 and zarr ?) (the extensibility mechanism is nice but difficult to find IMO because of hard to find or non existing documentation)

When I have to deal with hundreds of Sources, it's really a big plus to use `Spimdata` ( you can group sources by channel for instance). At the moment I have not found a better option and I already have some work which depend on it (including `bigdataviewer_bioformats`). 

But if you do not like the idea, then let's ignore `Spimdata` for the moment in this action. I will write the ones that I need. Anyway, `Spimdata` and  @hanslovsky's options or any other choice should be compatible with downstream actions acting on `Source`s. Just the input action is different.

tischi:
OK, as said I do not have much experience with `SpimData`. For reading and writing we will need it for sure. And I also do not know how BigStitcher works... I think that's something I really should look into! If you have time it would be nice to paste links to some code where `SpimData` is used in a useful way.

NicoKiaru:
Pfou... The singular stuff is not renamed in this branch and I still have the BdvUtils / BDVUtils issue. Even after deleting my fork and reimporting the project.

Is it too much work to redo this refactoring from the master branch ? Honestly I reached my max capabilities in git knitting here.

NicoKiaru:
Ah looks like I managed. I had to remove the extra BDVUtils file in the refactorSourceAdder branch

NicoKiaru:
How do you test your `Command` @tischi ? Do you have a main (test) method which initializes an IJ instance somewhere ?

NicoKiaru:
I've added a lot of things (but that remains into a single package) in order to use `BdvHandle` as SciJava parameters. However right now the autocontrast adjustement  is not working. This is because retrieving the index of the source is not working. The index source retrieval mechanism looks weird to me.

NicoKiaru:
Maybe we should push like this -> merging will enable to use and test the scijava parameter mechanism to master.

tischi:
I would be fine with just merging it and fixing the issues from the new master branch. As we do not require a working master branch at this stage I think this is easier.
What do you think?

NicoKiaru:
perfect!

PR #35, opened 2019-12-16
Fix bug in getSourceIndex() and rename brightnessAdjuster

tischi:
The Brightness adjuster works now for the mri-stack example image. I added a Demo.

But I could not find your example data in test resources, did you forget to _git add_ them?

### DIFF
--- a/src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
+++ b/src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
@@ -247,12 +247,22 @@ public class BdvUtils {
         System.out.println("look for : "+  source);
 
         for ( int i = 0; i < sources.size(); ++i ) {
-            System.out.println("c:"+sources.get(i).getSpimSource());
+            final Source< ? > bdvSource = sources.get( i ).getSpimSource();
 
-            final Source wrappedSource = ( ( TransformedSource ) sources.get( i ).getSpimSource() ).getWrappedSource();
+            System.out.println("c:"+ bdvSource );
 
-            if ( wrappedSource.equals(source) )
-                return i;
+            if ( bdvSource instanceof TransformedSource )
+            {
+                final Source wrappedSource = ( ( TransformedSource ) bdvSource ).getWrappedSource();
+
+                if ( wrappedSource.equals( source ) )
+                    return i;
+            }
+            else
+            {
+                if ( bdvSource.equals( source ) )
+                    return i;
+            }
         }
 
         return -1;
--- a/src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
+++ b/src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
@@ -1,6 +1,7 @@
 package sc.fiji.bdvpg.bdv;
 
 import bdv.tools.brightness.MinMaxGroup;
+import bdv.tools.transformation.TransformedSource;
 import bdv.util.Bdv;
 import bdv.util.BdvHandle;
 import bdv.viewer.Source;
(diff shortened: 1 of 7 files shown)


NicoKiaru:
There are no data because the sources are generated by the code, see command:

https://github.com/haesleinhuepf/bigdataviewer-playground/blob/master/src/main/java/sc/fiji/bdvpg/command/BdvAppendSampleCommand.java

The auto contrast adjust is disabled in the command. ( and so you don't see anything in bdv unless you put the max value to 256). I will try it with your fix.

NicoKiaru (commented):
In BdvUtils, the logic to find the Source index makes assumptions : it's that the source you are looking for is a wrapped source. This may not be always the case. At least a test with `if (src instanceof TransformSourced)` could be added. And also if no index is found, maybe an error message should be written ? Or returning -1 is enough?

NicoKiaru:
Otherwise : I'm impressed that the autocontrast works for the generative sources (wave3D and mandelbrot) However, it's failing with the more basic voronoi sample (small and big). Do you think it can be fixed ? I do not know at all the logic behind the autocontrast, but does it depend on the bdv window location ?

tischi:
> In BdvUtils, the logic to find the Source index makes assumptions : it's that the source you are looking for is a wrapped source. At least a test with if (src instanceof TransformSourced) could be added. 

- [x] Yes, I will add this test, however the current design of Bdv is such that it always is a `TransformedSource`, I think...

> And also if no index is found, maybe an error message should be written ? Or returning -1 is enough?

This whole issue might disappear once Tobias got rid of the indexing....
For now, I think we could even throw an error, unless there would be scenario's where this could happen? @NicoKiaru ok?!

> However, it's failing with the more basic voronoi sample (small and big)

- [x] I will have a look!

NicoKiaru:
>  I will add this test, however the current design of Bdv is such that it always is a TransformedSource, I think...

Yes,  but also we could wrap it several times... Let's keep it the logic like that for the moment, and we'll deal with this theoretical situation when it happens for real.

tischi:
> Yes, but also we could wrap it several times... Let's keep it the logic like that for the moment, and we'll deal with this theoretical situation when it happens for real.

I know! In fact, for myself, I had very complex code there, unwrapping tons of stuff. The problem is: We cannot know how people will wrap their sources and we might not even know all the classes that could be involved. Thus, long term, I think we need a different logic.

NicoKiaru:
For the record, what I implemented to inspect a Source :

https://github.com/BIOP/bigdataviewer_scijava/blob/1c5c1e9fc06de0e93ccc8cb30de54a995ad2cc78/src/main/java/ch/epfl/biop/bdv/scijava/command/BdvSourcesInspect.java

Maybe you can add yours and we'll find a way to add this in the playground

tischi:
@NicoKiaru 

- I added the check for the `TransformedSource`.
- It does not work for the Voronoi, because it currently only works for `UnsignedShortType`. I do not have time to fix this right now. I will make an issue and suggest you merge this PR anyway.

NicoKiaru:
Great! Thanks for looking!

tischi:
> For the record, what I implemented to inspect a Source ...

As said I think we cannot know what people will do, so we cannot cover all the situations. Let's wait what Tobias comes up with during his redesign.
Could you please make an issue about this?

NicoKiaru:
You mean an issue about the Source selection ? I haven't checked at all how it's done, so maybe you are better fitted to make an issue about this ?

But I can make an issue about the Brightness Contract for voronoi, if that's what you meant.

Let me know

tischi:
> But I can make an issue about the Brightness Contract for voronoi, if that's what you meant.
I did that already

> You mean an issue about the Source selection
Yes, I can do it!

NicoKiaru:
Yes thank you!

PR #39, opened 2019-12-17
Newbigwarp

NicoKiaru:
Hi @tischi and @haesleinhuepf , 

Here's an implementation of a BigWarp launcher.

Let me know what you think. Small demo video here :     
https://youtu.be/24p4pK83eXo


NicoKiaru:
Technical details:
* For the [`Command`](https://github.com/haesleinhuepf/bigdataviewer-playground/blob/f1a031af0a9e5bf43d3975812ce55792f74504c6/src/main/java/sc/fiji/bdvpg/command/BigWarpCommand.java), the input parameters are:
  * a `BdvHandle` and indexes of moving sources, as a `String` 
  * a `BdvHandle` and indexes of fixed sources, as a `String`

The BdvHandle can be the same for moving and fixed source, but it's convenient to keep the possibility to have them separated.

Output:
* The 2 `BdvHandle` generated by BigWarp. Because they are declared as `ItemIO.OUTPUT`, a companion JPanel containing the list of Sources present in the BdvHandle is displayed.

To retrieve the transform from `BigWarp`, what can be done is call the `getTransform()` function from any `WarpedSource` present in BigWarp windows (source of index 0 is fine).

NicoKiaru:
Also, what's fed into `BigWarp` is `SourceAndConverter`:

https://github.com/haesleinhuepf/bigdataviewer-playground/blob/f1a031af0a9e5bf43d3975812ce55792f74504c6/src/main/java/sc/fiji/bdvpg/source/register/BigWarpLauncher.java#L90

This allows to transfer the `B&C` and `Color` of sources. One drawback is that the `ConverterSetup`is not transfered : if you type `s` the sliders and color selection for the source will not be displayed...

haesleinhuepf:
I love it! Just a minor request: Could we limit the package depth? When it's too deep it becomes hard to explore.

I suggest to change
```
/sc/fiji/bdvpg/bdv/source/get/
```
to
```
sc/fiji/bdvpg/bdv/source
```

What do you think?

Cheers,
Robert

tischi:
@haesleinhuepf I would fully trust you in terms of package structure. What about you make a PR where you change everything such that you think it makes sense?

NicoKiaru:
Thanks for the feedback. I'll push and then let's restructure the package.

PR #40, opened 2019-12-17
Source viewer

tischi:
Hey guys,

I added a `SourceViewer` class, see demo here: https://github.com/haesleinhuepf/bigdataviewer-playground/blob/dba1c082a0449f5fea9c1937df64351d60436771/src/test/src/sc/fiji/bdvpg/bdv/source/view/SourceViewerDemo.java#L19

I think this would allow us to get rid of the `BDVSingleton` stuff, wouldn't it?


NicoKiaru:
Actually with the SciJava stuff, the BDVSingleton is not used.

I'm not quite sure to understand the use of the SourceViewer, but I'm probably missing something.

tischi:
> I'm not quite sure to understand the use of the SourceViewer, but I'm probably missing something.

In contrast to all our other functions, the `SourceViewer` does not take a `bdvHandle` as an input but provides one as an `output`. So, it is a means to create a new `bdvHandle`.

NicoKiaru:
Ok, but then why do we need to give a `Source` ? Could we just give no argument and get an empty `BdvHandle` as an output ? Otherwise, its really similar to `BdvFunctions.show`, no ?

tischi:
it is not only similar, it is the same :-)
but we have many such short convenience wrappers.
there is no such thing as an empty bdv afaik, there needs to be an empty dummy image, which can be very confusing later.

NicoKiaru:
> such thing as an empty bdv afaik

One trick to do it : 

see https://github.com/haesleinhuepf/bigdataviewer-playground/blob/addcfe367d0c494fc13a33b38e1ee612606f5c36/src/main/java/sc/fiji/bdvpg/command/BdvWindowCreateCommand.java#L56-L66

Create one with a dummy image then remove it ;-)

tischi:
ok, I did not know this :-)
Shall we still have the SourceViewer class?

NicoKiaru:
Yes, but maybe just without any `Source` argument, and renamed ? 
Something like `SourceViewerSupplier` or `BdvHandleSupplier` with a `run` method and a `get` method that returns an empty BdvHandle ?

We can keep the hack, and when we have a better way of making an empty bdv, we'll remove the hack. How does that sound for you ?

tischi:
ok. we can go with the empty hack.
Then I remove this branch?!

NicoKiaru:
You can rename, put the hack, and remove the source input in your `SourceViewer` and then we merge it. Sounds good ? I will then change my `BdvCreateWindow` command to use this action because right now it's not an action, and it's not the design we chose. So I'll fix it (after the merge).

tischi:
@NicoKiaru 
Ok, I tried to do what you suggested. Can you check whether I understood you correctly?

NicoKiaru:
Wonderful! I just modified the Command that was creating the BdvHandle to use the BdvCreator action.

NicoKiaru (approved):
All good for me.

PR #45, opened 2019-12-18
Fix bug in viewer voxel spacing computation

tischi:
- Fixed bug in viewer voxel spacing computation (replaced whole method)
- Some small refactoring (just gardening)


NicoKiaru:
Conflicts fixed -> just an issue in the action package location.

Issue #32, opened 2019-12-15
Major refactoring suggestion


NicoKiaru:
Hi @tischi and @haesleinhuepf, 

I have a suggestion to make about the structure of the repo. What do you think of restructuring like this:
* `bdv` : A BdvHandle specific package containing all actions that strictly depend on a `BdvHandle`:
  * Root : Bdv creation
  * Root : Bdv closing
  * `bdv.navigate` : Loging, getting current position, synchronizing position of `BdvHandle`(s) 
  * `bdv.group` : All actions related to Source grouping
  * `bdv.source` : actions on Sources which depends on the presence of a `BdvHandle`
    * `bdv.source.screenshot` : Screenshot action
    * `bdv.source.displayopts` : source display options : (B&C, hide, show, set Color) 
    * `bdv.source.append` : Action which adds Sources to `BdvHandle` (append a single `Source`, or a `List<Source>`, or a `SpimData`, or a `SourceAndConverter` etc)
    * `bdv.source.get`: action which outputs `Source`s in various ways from a `BdvHandle` (principally from source indexes for instance) or `SourceAndConverter`

* `source`: Actions on Sources, as much independent of `BdvHandle` as possible
  * `source.transform`: Affine transformation / Warping transformation, Resampling etc
  * `source.register`: here would go automated registrations and BigWarp (BigWarp is BdvHandle dependent however)
    * `source.samples` : get sample Sources
    * `source.import` : import `List<Sources>` from Xml/Hdf5 , or from a tiff, or from IJ1 with the active window...
    * `source.export` : export List<Source> into a dataset, or a ome.tiff file, etc.

As discussed with @tischer I would need to use `Spimdata` so a specific `Spimdata` package:

* `spimdata`
  * `spimdata.import` get `SpimData` from Xml/Hdf5 or from BigdataServer
  * `spimdata.export` basically save an open spimdata object

tischi:
love it, do it!

NicoKiaru:
Cool!

One thing where I'm not completely sure is where to put the Commands:
* they are not as granular as these actions -> so we will not have one ui per action
* they will combine actions placed in several of these packages.

Here are several options:
* 1 We can put commands in their 'supposed' main package
* 2 We separate commands fully from actions:
  * maybe the package substructure can match the location in the FIJI Menu

What do you think @tischi and @haesleinhuepf ?

tischi:
[...]

PR #48, opened 2019-12-20
SourceAndConverter Demo

tischi:
Hi @NicoKiaru,

This branch now depends on the PR being merged.

You could have a look here how that would work:
https://github.com/bigdataviewer/bigdataviewer-playground/blob/6742fdeeb9369e3423f24aad7130f378ecdc73a5/src/test/src/sc/fiji/bdvpg/bdv/sourceAndConverter/SourceAndConverterLoaderDemo.java#L12

If we go from there we could for example explore having global maps:

- Map< Source, SourceAndConverter > [EDIT: We do not need this because the sac can be obtained from the bdvss]
- **Map< Source, BdvStackSource >**

and explore how that would work.


NicoKiaru:
Very cool! However your action needs to replace the action `AddSourceAndConverterToBdv` which is currently in the package `sc.fiji.bdvpg.bdv.source.append`: it's doing the same action, but is obsolete now.

`AddSourceAndConverterToBdv` is an action which I use for the BigWarpLauncher. Modifiying it will allow to test if everything works smoothly.

And moving the `SourceAndConverterAdder ` to `sc.fiji.bdvpg.bdv.source.append` goes according to the rules set in the package structure design decided in https://github.com/bigdataviewer/bigdataviewer-playground/issues/32

I can do it on Monday.

NicoKiaru:
I did the refactoring. Your demo works. 
I tried BigWarpLauncher and I got this error:
```
[ERROR] Module threw error
java.lang.NoClassDefFoundError: net/imglib2/display/RealARGBColorConverter$Imp0
	at bigwarp.BigWarp.addWarpMagnitudeSource(BigWarp.java:1646)
	at bigwarp.BigWarp.<init>(BigWarp.java:335)
	at bigwarp.BigWarp.<init>(BigWarp.java:293)
	at sc.fiji.bdvpg.source.register.BigWarpLauncher.run(BigWarpLauncher.java:115)
	at sc.fiji.bdvpg.command.BigWarpCommand.run(BigWarpCommand.java:57)
	at org.scijava.command.CommandModule.run(CommandModule.java:199)
	at org.scijava.module.ModuleRunner.run(ModuleRunner.java:168)
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:127)
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:66)
	at org.scijava.thread.DefaultThreadService$3.call(DefaultThreadService.java:238)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
Did you get this error ?
Maybe it's a version issue with ImgLib2.

tischi:
@NicoKiaru

## Versions

I also had issues with versions.  
I had to specify the version of `bigdataviewer-core` to make things run.

Maybe it is a good thing that we are more specific about versions in our pom at this stage, because we might use newer stuff than what Fiji (scijava parent-pom) is currently using.

You could try fixing the imglib2 version to the latest one.

https://maven.scijava.org

```
<dependency>
  <groupId>net.imglib2</groupId>
  <artifactId>imglib2</artifactId>
  <version>5.8.0</version>
</dependency>
```

Also, if you add a `BigWarpLauncherDemo` I can test it for you and, hopefully, figure out the issue.

## Package structure

Sorry, I did not pay much attention! Please feel free to reorganise!

## General

I am glad you like it! 
Working more with `SourceAndConverter` and `BdvStackSource` seems to me the most promising approach in order to improve our playground :-)
It would be really nice to have Tobias' input here, though.
Maybe you could also try poke him on imglib2/gitter at some point about this.

NicoKiaru:
Done for the demo. Let me know if you manage to solve the issue. I will work again tomorrow on the playground and then will go for the vacation break.

NicoKiaru:
I increased the version of BigWarp - I got a different error because BigWarp is expected a VoxelDimension, which the sample source do not. I filed an issue on BigWarp (https://github.com/saalfeldlab/bigwarp/issues/83) but I will also try to add VoxelDimension to the sample source to go through this error

NicoKiaru:
@tischer 
Ok, so I added a `VoxelDimension`to mandelbrot. That was unnecessarily painful. This fixes partly the issue. Adding a voxeldimension to mandelbrot is also a bit annoying because there is no constructor of `RandomAccessibleIntervalSource` which allows to specifiy a `VoxelDimension`. I did not do it and thus in the demo I 'bigwarp' mandelbrot with itself.

But anyway, this is enough to see one thing: when you want to open BigWarp, specifying the `SourceAndConverter` do not transfer the `ConverterSetup` of the initial `Source`. To notice this, press `S` in the BigWarp demo instance : it is not possible to set the color and min max of the moving and fixed source.

That's because the ConverterSetup is not attached to the SourceAndConverter.

We need to think carefully before going further.

I like the idea of working with `SourceAndConverter`, but I do not like so much working with `BdvStackSource` : contrary to our design, it strongly links a `Source` to a single `BdvHandle`.

In our discussion with @haesleinhuepf , we thought that it was a good idea to have somewhere a map linking `Source` (or maybe `SourceAndConverter`) to useful "metadata" in a broad sense, but independently of BdvHandle (it can be displayed multiple times). 

From BdvStackSource, I'd rather keep the method that set mix and max and the method that sets the color ( so a `ConverterSetup` object). Otherwise, I would just drop the rest. This would probably also allow us to continue without depending on PRing Bdv-vistools.

Let's discuss this when you have time.

tischi:
> From BdvStackSource, I'd rather keep the method that set mix and max and the method that sets the color ( so a ConverterSetup object). 

I had a look and agree. If we have the `ConverterSetup` we could refactor all functionality in `BdvStackSource` into small classes that are consistent with our current design (and thus would not need to use `BdvStackSource`). I like that better as well.

> This would probably also allow us to continue without depending on PRing Bdv-vistools.

Here I am not so sure, because how can we construct a functional `ConverterSetup` without `BdvFunctions.show( SourceAndConverter )`? (Unless, see "General thoughts" below...)

> But anyway, this is enough to see one thing: when you want to open BigWarp, specifying the SourceAndConverter do not transfer the ConverterSetup of the initial Source. To notice this, press S in the BigWarp demo instance : it is not possible to set the color and min max of the moving and fixed source.

I actually find this logical, because a `SourceAndConverter` has no `ConverterSetup`. I think a `ConverterSetup` is an object that essentially wraps a `ColorConverter` and exposes control of its brightness settings.

### General thoughts

If we add a `SourceAndConverter`, if `converter instanceof ColorConverter` it has all the functionality we need to modify it's brightness, so it seems we would not need a `ConverterSetup` at all. And if `! converter instanceof ColorConverter` one cannot obviously modify the brightness. In that case we could do the same that `bdv-vistools` is currently doing, i.e. wrap a `ColorConverter` around the `Converter` (I can explain you...).

**Thus seems like to me that the main decision now is whether we want to rely on (be compatible with) the current and future (@tibuch) UI of bdv in terms of setting the brightness of sources.** If we do not depend on using the "official" UI we can probably simply add a `SourceAndConverter` via the `viewerPanel` and write our own UI components around it. I think we both have enough code here such that it would not be a big deal and it would be faster to progress, but still we would significantly branch off here. 

What do you think?

NicoKiaru:
> Thus seems like to me that the main decision now is whether we want to rely on (be compatible with) the current and future (@tibuch) UI of bdv in terms of setting the brightness of sources.

It looks like we are going a parallel way for the moment, and this helps us to go forward. We can merge our work once we know more what we be released.

NicoKiaru:
> I had a look and agree. If we have the ConverterSetup we could refactor all functionality in BdvStackSource into small classes that are consistent with our current design (and thus would not need to use BdvStackSource). I like that better as well.

👍 

> If we add a SourceAndConverter, if converter instanceof ColorConverter it has all the functionality we need to modify it's brightness, so it seems we would not need a ConverterSetup at all. And if ! converter instanceof ColorConverter one cannot obviously modify the brightness. In that case we could do the same that bdv-vistools is currently doing, i.e. wrap a ColorConverter around the Converter (I can explain you...).

Regarding this, in the PR https://github.com/bigdataviewer/bigdataviewer-playground/pull/51, one way to achieve this is to build our own `ConverterSetup` object and link it to a Source. I believe having the `ConverterSetup` is still useful for the GUI. However, since we do our own sauce, we can wrap a  ColorConverter, or any other converter. In the PR, there is a working version of using LUT from IJ1 into bigdataviewer, which uses a custom `Converter` and a standard `ConverterSetup` object.

PR #49, opened 2019-12-29
SourceAndConverter v2

tischi:
@NicoKiaru 

I started making a version without the `BdvStackSource`. 

See here for a working example:
https://github.com/bigdataviewer/bigdataviewer-playground/blob/f84da2cbf4a1595dfb10e48f45a3a7aa94ee2384/src/test/src/sc/fiji/bdvpg/bdv/sourceAndConverter/SourceAndConverterLoaderDemo.java#L15

What do you think?


src/main/java/sc/fiji/bdvpg/bdv/source/append/SourceAndConverterBdvAdder.java
```diff
@@ -1,38 +1,26 @@
...
     }
 
     @Override
-    public BdvStackSource apply(SourceAndConverter sourceAndConverter) {
-        int numTimePoints = 1;
-        // IMO BdvFunctions.show should take care of the timepoints...
-        final ConverterSetup converterSetup = bdvHandle.getSetupAssignments().getConverterSetups().get( 0 );
-        return BdvFunctions.show( sourceAndConverter, numTimePoints, BdvOptions.options().addTo(bdvHandle) );
+    public void accept(SourceAndConverter sourceAndConverter) {
+        bdvHandle.getViewerPanel().addSource( sourceAndConverter );
```
NicoKiaru, line 24: ok, so here you drop the convertersetup, right ?

src/main/java/sc/fiji/bdvpg/sourceandconverter/display/BrightnessAdjuster.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.function.Consumer;
+
+public class BrightnessAdjuster implements Runnable, Consumer< SourceAndConverter > {
+
+    BdvHandle bdvHandle;
+    SourceAndConverter< ? > sac;
+    double min;
+    double max;
+
+    public BrightnessAdjuster( final BdvHandle bdvHandle, final SourceAndConverter< ? > sac, double min, double max )
```
NicoKiaru, line 17: Because the convertersetup is not passed to bdv, one has to call this action to change brightness and contrast. I'm ok with that, but then it means we need to provide a new convenient ui.

Also one remark : if a source and converter is displayed in multiple bdvhandle, one need to call this action as many times as there are bdv windows displaying this sourceandconverter object.

src/test/src/sc/fiji/bdvpg/bdv/sourceAndConverter/SourceAndConverterLoaderDemo.java
```diff
@@ -1,37 +1,34 @@
...
-		// add
-		//
-		final SourceAndConverterBdvAdder adder = new SourceAndConverterBdvAdder( bdvHandle, sourceAndConverter );
-		adder.run();
-
-		// change brightness
-		//
-		final BdvStackSource< ? > bdvStackSource = adder.getBdvStackSource();
-		bdvStackSource.setDisplayRange( 0, 255 );
+		return loader.getSourceAndConverter( 0 );
```
NicoKiaru: The output are supposed to not have any parameter according to our design. Returning a List of all source and converter could be better, also it avoids calling BigDataViewer.init as many times as there are sources you want to get.

NicoKiaru (commented):
I think it's good to try to avoid using BdvVisTools, and thus to deal directly with SourceAndConverter. However there are several limitations in the current design which I think will be pretty annoying if we continue along this path:
* Mainly, the current design works well if only one BdvHandle is used. If you have a Source displayed in two windows, you need to call the B&C command twice. And how to keep track of all Bdv Windows which are displaying a Source ? 
* Currently the B&C command for SourceAndConverter do not use  ConverterSetup, the object which makes it easy for the user to change color and B&C. We need to solve this, either by making a new convenient ui (using DynamicCommand ?) or by finding a way to reintroduce ConverterSetup. But again, how to make this easy if multiple BdvHandle are displaying the same source ?

I feel that we suffer somehow from the same problem IJ1 had : the data is too much link to the display. Having to deal with BdvHandle when we want to access the data is problematic (why getting the display to access the data ?). And if we find workarounds, they may be tedious to maintain on the long term.

My thoughts:
* We need to have a way to access the data (= Source) independently of the display. Like a Set of Sources, or better, a Map where keys are Source and values are a set of useful objects linked to the Source ( Volatile Source, Converter, ConverterSetup ... )
* We need to have a way to modify the display, which is somehow agnostic of the number of Bdv Windows opens, and which relies on the source Map to retrieve, when needed, the Volatile Source, Converter and ConverterSetup. Also the display handler would need to have functions which are somehow agnostic of the number of BdvHandle opened, so that when a source display option is modified, it can act on all bdv windows displaying the source.

One way to do this is by using scijava services. I have something working which I'll push next week. But if we want to avoid scijava services, there are ways to do this, but I believe the points mentioned above remain valid.

NicoKiaru:
But maybe this was not the scope of the repo ? I mean we said atomic action need a BdvHandle as a parameter, but already when we separated the repo into actions acting on source vs actions acting on bdvhandle we broke this design (https://github.com/bigdataviewer/bigdataviewer-playground/issues/32).

If we now want to clearly separate display and data, that may go a step further. (Needed in my opinion, but what do you think ? @tischi @haesleinhuepf )

PR #50, opened 2019-12-30
Implement different projection modes

tischi:
I implemented a demonstration of an average projection (rather than the default sum projection).
I urgently need this for our EM projects.

### Notes

This PR is based on (i.e. includes): https://github.com/bigdataviewer/bigdataviewer-playground/pull/49

### Example code 

https://github.com/bigdataviewer/bigdataviewer-playground/blob/ba217040fc4e96b41b646a7dca15ca51e05b6930/src/test/src/sc/fiji/bdvpg/bdv/sourceAndConverter/ProjectorDemo.java#L19

### Example image

Note how the **intensity is not increasing but averaged** where the two heads overlap:
image

### Next steps

For now I just hardcoded this behaviour, but an important question that we need to tackle is how to store various projection modes of our sources.

- [ ] Until we come up with a better idea I will implement a global `WeakHashMap<Source, ProjectionMode>` and let the `AccumulateProjector` act according to this map.


NicoKiaru:
Looks super cool and useful, but shouldn't AccumulatorProjectors be a property of a BdvHandle rather than of a Source ? So needing a WeakHashMap from BdvHandle to AccumulateProjector ?

Issue #47, opened 2019-12-19
Volatile and NonVolatile Sources


tischi:
Hey @NicoKiaru ,

The way our code works right now in terms of `SourceLoader` and `SourceAdder`: it would in fact show the nonVolatile `Source`, right? 

If that's the case the current logic would actually kind of be wrong...

So, what we actually want to show is the `SourceAndConverter`, right?
So `SourceAdder( SourceAndConverter sac )`, right?

NicoKiaru:
> The way our code works right now in terms of `SourceLoader` and `SourceAdder`: it would in fact show the nonVolatile `Source`, right?

I think so (unless you give a Volatile Source... but then it's an issue also).

> So, what we actually want to show is the SourceAndConverter, right?

Yes, but can we always do that ?
There is already a simple action to do that which I use for BigWarpLauncher (https://github.com/bigdataviewer/bigdataviewer-playground/blob/master/src/main/java/sc/fiji/bdvpg/bdv/source/append/AddSourceAndConverterToBdv.java). (It needs to be renamed)

tischi:
Yes, but doing it via the `ViewerPanel`

1. we do not get a `BdvStackSource` back, which I think is a real issue, because it might well be that this could be an important object for several things in our library.
1. did not add it to the BDV UI when I tried it, which I also think is a real issue, because we would likely not be able to benefit from @tibuch's new UI, once it is ready.

Thus, for me the way to proceed would be to get this PR merged: https://github.com/bigdataviewer/bigdataviewer-vistools/pull/40

I would even consider basing further developments on our own branch of bigdataviewer-vistools until it is merged. The point being that after our discussion yesterday I think `SourceAndConverter` and `BdvStackSource` might be key objects to work with in our playground, and, to me, the by far most natural entry point to this appears to me  `bdvss =  BdvFunctions.show( sac )`.

What do you think?

NicoKiaru:
> 1. we do not get a `BdvStackSource` back, which I think is a real issue, because it might well be that this could be an important object for several things in our library.

Right

> 2. doing it like this does not add it to the BDV UI, which I also think is a real issue, because we would likely not be able to benefit from @tibuch's new UI, once it is ready.

Right

So we need to somehow implement your bdvvistools pr (https://github.com/bigdataviewer/bigdataviewer-playground/issues/29#issuecomment-567529812) to fix this

tischi:
[...]

Issue #41, opened 2019-12-17
getSourceIndex( bdvHandle, source )


tischi:
This function: https://github.com/haesleinhuepf/bigdataviewer-playground/blob/addcfe367d0c494fc13a33b38e1ee612606f5c36/src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java#L242

...is brittle because it assumes that we have the `Source` that `bdv` holds internally as an object, or that we know how `bdv` could have wrapped or modified it internally such that we can still identify it. It feels like that we should do something entirely different here. 

For example, one could start looking where this function is used and see what it would take to avoid calling it at all.

NicoKiaru:
It's definitely a big issue... I guess it's the sort of issue we'll need to figure with Tobias if we do not want to write too many useless things.

Personnally, I'd say it would be better to do this completely differently:
* basically we would need a holder for all the Sources we are working on, and then we would have a function which returns all the `BdvHandle` which are displaying any `Source`. So basically a function `List<BdvHandle> getBdvHandlesDisplayingSource(source); `. 
* Then all the commands would have `@Parameter` annotations for the `Source`s being processed (or `SourceAndConverter` maybe...), and none for `BdvHandle`. Then a modification to a source would trigger a repaint action to the potentially multiple `BdvHandle`s which are displaying this source...

tischi:
so maybe we do ask Tobias to join the Skype tomorrow?

NicoKiaru:
Yep let's try ? It would be good to have an idea of what him and Tim-Oliver are planning.

Do you think you have time to review the BigWarp launch  (https://github.com/haesleinhuepf/bigdataviewer-playground/pull/39) before the call tomorrow ? I think it would be good to know what you think of this design, potentially you and Robert.

tischi:
I don't have time. I have to give a talk tomorrow...

NicoKiaru:
Ok, good luck for the talk @tischer! Maybe @haesleinhuepf , a quick look of #39 before the skype call ?

NicoKiaru:
Indexing logic hidden inside a display service, which takes care of it.

https://github.com/bigdataviewer/bigdataviewer-playground/pull/53

Issue #29, opened 2019-12-15
Add SourceAndConverter to Bdv


tischi:
@NicoKiaru 

Wow.....

https://github.com/haesleinhuepf/bigdataviewer-playground/blob/cdcad0549f9284b2e76f2f86fdd385e4116c5dac/src/main/java/sc/fiji/bdv/sourceAndConverter/add/AddSourceAndConverterToBdv.java#L22

I did not know this function existed for a `SourceAndConverter`!

1. What is the difference to `BdvFunctions.show( ... , BdvOptions.options.addTo( ... ) )`
1. Does it also add the DisplaySettings for this Source to the Bdv UI?

NicoKiaru:
Not available in `BdvFunctions` but in `ViewerPanel`, yes. So, it's accessible from `BdvHandle`. That's want I wanted to tell you but I wasn't clear enough when I told you to use `SourceAndConverter` instead of `Source`

That's the trick, this function allows to add a `SourceAndConverter`:
* It deals with `VolatileViews` because the `SourceAndConverter` object has a `asVolatile()` method which, if not null, contains the volatile representation of the Source.
* one issue : the `Converter` stays linked to the initial `ConverterSetups` -> if you add two identical `SourceAndConverter`, a single `ConverterSetup` which has a single slider will control the minmax of these two Sources. This can be positive or negative, depends.

NicoKiaru:
But wait for it : here's the magic of Functional interface:

NicoKiaru:
Let's say you want to affine transform a `SourceAndConverter`:
This means you need to apply an affineTransform on both the concrete and Volatile source, if present.
So one way is to create an affinetransform action for SourceAndConverter.

Then if you want to warp a Source, it's the same, you write an action for `Source` objects but also for `SourceAndConverter`objects. 

Then if you have an action which applies a Gaussian Blur it's the same. 

At the end you have many duplicated Actions, where you do the same things.

But ahah! All these actions on sources (affine, warp, gaussian, whatever) all implement `Function<Source,Source>` so what you need to do is just to call an action which will take any of these functions and apply it on a `SourceAndConverter` object.

And it's already done: look at:

https://github.com/haesleinhuepf/bigdataviewer-playground/blob/cdcad0549f9284b2e76f2f86fdd385e4116c5dac/src/main/java/sc/fiji/bdv/sourceAndConverter/SourceAndConverterApplySourceFunction.java

[...]

PR #51, opened 2020-01-05
Scijavaservice

NicoKiaru:
Hi @tischi and @haesleinhuepf, 

So here's my suggestion branch to solve several issues.
The idea is to have:
* a `BdvSourceService` which holds a List of Source, independently of their display. It maintains a Map which uses the Source as key. This map can contains as many properties as needed which should be attached to the Source. The most important and expected objects : a `Volatile` view, a `Converter`, and a `ConverterSetup`. But you can add anything which could be needed (metadata, spimdata, or a projector method as the one @tischi was needing).
* a second service `BdvSourceDisplayService`, which is responsible for everything display related. It should be responsible for building on the fly the volatile source, converters, etc. This service uses the BdvSourceService to hold these objects. 

For now I've build these two services as `SciJava` services, because honestly that was the fastest and most convenient way with what I know. But I know that you may want to skip these, so I've made two interfaces that can be replaced by an implementation you would prefer, if this approach seems promising to you. 

Issues that should be solved using this approach:
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/47
How ? A bdvSourceService, maintains a Source reference, independent of the display. A BdvSourceDisplayService, that uses a BdvSourceService, stores a single volatile view, a Converter and a ConverterSetup, by using the Source as a key. When a Source is displayed, it constructs a SourceAndConverter on the fly by fetching (or constructing if necessary) the appropriate SourceAndConverter object.
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/41
There is no indexing anymore needed for a Source. Actually some of the indexing logic is hidden in the BdvSourceDisplayService. So if the indexing logic completely disappear, it will just need to be updated within the service.
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/29

Already working:
* if a `Source` is displayed multiple times, the `ConverterSetup` reacts and updates all the Bdv windows where the `Source` is displayed
* `BigWarp` is working nicely 
* the projector branch is merged, but I'm not sure it's working as expected currently. Can you check and fix this @tischi ?
* all previously written actions are working

Things that should work next but which are not currently implemented:
* the volatile view wrapping is not done yet for all sources
* there is no need for BdvVisTools, because we redo most of it. So if we need to add your own stuff, it will be easier
* we cannot remove sources yet, but the design should allow without too much suffering
* we could set easily a LUT converter, different from a single color one

 One annyoing issue : handling of the number of timepoints, because there's no direct way to know the number of timepoints form a `Source` object.


src/main/java/sc/fiji/bdvpg/scijava/processors/BdvHandlePostprocessor.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void process(Module module) {
+
+        module.getOutputs().forEach((name, object)-> {
+            if (object instanceof BdvHandle) {
+                BdvHandle bdvh = (BdvHandle) object;
+                log.accept("BdvHandle found.");
+                //------------ Register BdvHandle in ObjectService
+                os.addObject(bdvh);
```
imagejan, line 45: Instead of calling `ObjectService#addObject()`, you could also publish an [`ObjectCreatedEvent`](https://javadoc.scijava.org/SciJava/index.html?org/scijava/object/event/ObjectCreatedEvent.html), which leads to automatic registration with the `ObjectService`.

Then `BdvSourceService` and `BdvSourceDisplayService` could just as well listen to these `ObjectEvent`s and react accordingly.

I'm unsure what would be the best strategy. Maybe @ctrueden has some better advice.

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvSourcePostprocessor.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.function.Consumer;
+
+@Plugin(type = PostprocessorPlugin.class)
+public class BdvSourcePostprocessor extends AbstractPostprocessorPlugin {
+
+    @Parameter
+    BdvSourceService bss;
+
+    public static Consumer<String> log = (str) -> System.out.println(BdvSourcePostprocessor.class.getSimpleName()+":"+str);
```
imagejan, line 20: That's an interesting approach to logging. Why don't you use the SciJava `LogService` here?

pom.xml
```diff
@@ -113,7 +124,15 @@
...
             <groupId>sc.fiji</groupId>
             <artifactId>bigwarp_fiji</artifactId>
+
+			<!-- See Issue https://github.com/saalfeldlab/bigwarp/issues/83
+			<version>5.1.0</version>-->
         </dependency>
+		<dependency>
+			<groupId>net.imglib2</groupId>
+			<artifactId>imglib2</artifactId>
+			<version>5.8.0</version>
```
imagejan: For best compatibility with other projects in the SciJava ecosystem, please don't override the version like this.

Instead, change the version property defined in the `<properties>` section:

    <imglib2.version>5.8.0</imglib2.version>

---

The current `master` branch of `pom-scijava` already defines this version, so the version declaration here can be removed once `pom-scijava-28.0.0` is released. See [`pom-scijava/pom.xml#L340-L341`](https://github.com/scijava/pom-scijava/blob/a78ae8631fccaba7791c3b918c6f939dffab0b65/pom.xml#L340-L341)

imagejan (commented):
I stumbled upon this pull request and left a few comments 😄 
Thanks all for your collaborative work on this project!

pom.xml
NicoKiaru: Is there a way to do this for `<artifactId>scijava-ui-swing</artifactId>` as well ?

And thanks a lot for your comments, @imagejan !

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvSourcePostprocessor.java
NicoKiaru: We are still not sure to which point we would like to use Scijava service, so we try to keep the dependencies minimal. This PR uses quite a lot the Scijava framework, because I find it convenient, but I'd like to keep open the option of not using it, if possible. Still an open discussion.

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvHandlePostprocessor.java
NicoKiaru: Ok, I don't remember how to use the `EventService`. Do you have a simple example somewhere ?

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvSourcePostprocessor.java
imagejan: Ah, alright. To be honest, I don't see any reason to *not* use SciJava wherever possible. You're writing `Command`s and `Plugin`s, so you depend on SciJava anyway. The `LogService` itself doesn't bring in any additional dependencies.

pom.xml
imagejan: Yes, it should be done for *all* versions of *all* artifacts managed by `pom-scijava`. The general pattern is `<{artifactId}.version>x.x.x</{artifactId}.version>`.

To see if a component is managed by `pom-scijava`, look in its [`pom.xml`](https://github.com/scijava/pom-scijava/blob/master/pom.xml).

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvHandlePostprocessor.java
imagejan: This might be of use:

https://github.com/scijava/scijava-common/blob/560ad964a7a1e859edd4dbe4fc48d7d9e09523f2/src/main/java/org/scijava/display/DefaultDisplayService.java#L211-L218

src/main/java/sc/fiji/bdvpg/scijava/processors/BdvSourcePostprocessor.java
haesleinhuepf: The idea was to have bdv-related functionality in classes which don't use sci-java or imagej. We can then later merge functionality easier into bdv-core if desired... 😉

pom.xml
NicoKiaru: Thanks, that's changed in https://github.com/bigdataviewer/bigdataviewer-playground/pull/51/commits/e0d91c9b061a726cfde4c96c74be220facada629

PR #66, opened 2020-01-13
Change some names and add a AccumulateMixedProjector

tischi:
@NicoKiaru 
Please have a look whether you agree. There are no significant changes. 
Most of it is just cosmetic, to make it more understandable for my personal taste.


src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
```diff
@@ -46,7 +46,7 @@ private void createEmptyBdv()
 
 		bdvHandle = bss.getBdvHandle();
 
-		//bdvHandle.getViewerPanel().setInterpolation( Interpolation.NLINEAR );
+		bdvHandle.getViewerPanel().setInterpolation( Interpolation.NLINEAR );
```
NicoKiaru: You want to keep the linear interpolation by default ? It's pretty slow sometimes. Maybe we can put this as a parameter in this action ?

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateAverageProjectorARGB.java
```diff
@@ -47,7 +45,6 @@ public AccumulateAverageProjectorARGB(
 			final ExecutorService executorService )
 	{
 		super( sourceProjectors, sourceScreenImages, target, numThreads, executorService );
-		this.sourceList = sources;
```
NicoKiaru, line 50: Based on this constructor, you cannot (should not?) modify the number of sources displayed in the bdv window ?

src/main/java/sc/fiji/bdvpg/scijava/services/BdvSourceAndConverterDisplayService.java
```diff
@@ -115,7 +102,7 @@ public BdvHandle getActiveBdv() {
                                 true,
                                 "is2D", false,
                                 "windowTitle", "Bdv",
-                                "projector", "Sum Projector").get().getOutput("bdvh");//*/
+                                "projector", ProjectionTypes.MIXED_PROJECTOR ).get().getOutput("bdvh");
```
NicoKiaru: So mixed projector by default ? Don't you think we can avoid specifying here the projector ? This will trigger a list for the user (and it will remember the last choice).

src/main/java/sc/fiji/bdvpg/scijava/services/BdvSourceAndConverterService.java
```diff
@@ -85,23 +85,23 @@
      * Map containing objects that are 1 to 1 linked to a Source
      * TODO : ask if it should contain a WeakReference to Source keys (Potential Memory leak ?)
      */
-    Map<SourceAndConverter, Map<String, Object>> data;
+    Map<SourceAndConverter, Map<String, Object>> sourceAndConverterToMetadata;
 
     /**
-     * Reserved key for the data map. data.get(sourceandconverter).get(SPIMDATAINFO)
```
NicoKiaru, line 91: I'd rather keep 'info' somewhere in the field name because this also contains a reference to the view setup id, and in the future maybe some other stuff (like the original file or a list of successive files)

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -0,0 +1,152 @@
...
+		if ( bAccu > 255 )
+			bAccu = 255;
+
+		target.set( ARGBType.rgba( rAccu, gAccu, bAccu, aAccu ) );
+
+	}
+
+	private String getProjectionMode( int sourceIndex, Map< SourceAndConverter, Map< String, Object > > sourceAndConverterToMetadata )
+	{
+		Source< ? > source = sourceList.get( sourceIndex );
```
NicoKiaru, line 132: ah! Maybe we need to think in terms of metadata which is transfered and metadata which is not transfered. Like when we duplicate a SourceAndConverter, should we transfer the metadata ?

src/main/java/sc/fiji/bdvpg/spimdata/importer/SpimDataFromXmlImporterAndRegisterer.java
```diff
@@ -1,44 +1,42 @@
...
 import mpicbg.spim.data.SpimDataException;
-import mpicbg.spim.data.XmlIoSpimData;
 import mpicbg.spim.data.generic.AbstractSpimData;
 import sc.fiji.bdvpg.services.BdvService;
 
 import java.io.File;
 import java.util.function.Function;
 
-public class SpimDataImporterXML implements Runnable, Function<File, AbstractSpimData> {
+public class SpimDataFromXmlImporterAndRegisterer implements Runnable, Function<File, AbstractSpimData> {
```
NicoKiaru: Registerer is a bit overkill. Also this would mean that we have to add this name in many actions. Maybe we can avoid this ?

src/test/src/sc/fiji/bdvpg/bdv/SpimDataDisplayDemo.java
```diff
@@ -0,0 +1,34 @@
...
+		BdvHandle bdvHandle = BdvService.getSourceAndConverterDisplayService().getActiveBdv();
+
+		// Import SpimData
+		new SpimDataFromXmlImporterAndRegisterer("src/test/resources/mri-stack.xml").run();
+		new SpimDataFromXmlImporterAndRegisterer("src/test/resources/mri-stack-shiftedX.xml").run();
+
+		// Show all SourceAndConverter associated with above SpimData
+		BdvService.getSourceAndConverterService().getSourceAndConverters().forEach( sac -> {
+			BdvService.getSourceAndConverterDisplayService().show(bdvHandle, sac);
+			new ViewerTransformAdjuster(bdvHandle, sac).run();
```
NicoKiaru, line 30: Adjusting the view on each source is not necessary, but adjusting the view on all sources is not an available action on the moment. Maybe we should implement that ?

NicoKiaru (commented):
Good! I have a few comment in the review, can you have a look at it ?

src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
tischi: done in https://github.com/bigdataviewer/bigdataviewer-playground/pull/66/commits/e1b70a8a3af1b4a67865f740f01a1abb3ce695b4

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateAverageProjectorARGB.java
tischi: No, they can be modified. In fact, this class is instantiated constantly by Bdv (I think for every new render thread).

src/main/java/sc/fiji/bdvpg/scijava/services/BdvSourceAndConverterDisplayService.java
tischi: done in https://github.com/bigdataviewer/bigdataviewer-playground/pull/66/commits/a8c8c4aab2c883d19ddfcf1a8994cf33ca4bbc84

src/main/java/sc/fiji/bdvpg/scijava/services/BdvSourceAndConverterService.java
tischi: done in https://github.com/bigdataviewer/bigdataviewer-playground/pull/66/commits/114d31a3b4b9b50365f50c2ab2f562c3b915ea00

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: I did not look into the duplicate method yet. Could we merge this for now and deal with this later?

src/main/java/sc/fiji/bdvpg/spimdata/importer/SpimDataFromXmlImporterAndRegisterer.java
tischi: Ok, I see. I changed it back for now in https://github.com/bigdataviewer/bigdataviewer-playground/pull/66/commits/4ef61ca423cd2e79db842dfa9d1d5f55f05a9e88

tischi: Maybe we have a smart idea later how to do this better.

src/test/src/sc/fiji/bdvpg/bdv/SpimDataDisplayDemo.java
tischi: Yes, we should! 
I made an issue: https://github.com/bigdataviewer/bigdataviewer-playground/issues/67

tischi (commented):
I think I addressed everything.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
NicoKiaru: sure!

PR #71, opened 2020-01-17
New Manual Registration

NicoKiaru:
This branch adds manual registration possibilities which is different from the manual transformation from the original bdv:
* It does not require that all sources are transformed sources

It temporarily creates Transformed SourceAndConverter during the manual registration, then, when the registration is done, the way the transformation is stored can be selected:
* Either the source is wrapped into another transformed source with the new transformation
* If the source was already a wrapped into a Transformed source, the transform can be mutated to account for the new transform
* If the source originates from a SpimSource, the last transform of spimdata can be mutated / or a new transform is added

There's a demo code to test all these outputs + there is a Command to execute these registrations with a UI, with the action registered into the popup window.

It's working, but here's a list of things to do / to make sure before merging: 

- [x] clean code

- [x] add documentation

- [x] make it cancelable

- [x] ensure the applied transformation keeps orthonormality

EDIT : 

I remove create a command to save spimdataset, I think there's enough in this branch for the moment.


tischi:
Hi @NicoKiaru 
I don't have much time tonight nor next week to thoroughly test it.
I glanced over it and it seems to mainly add functionality and not change anything fundamental.
I would thus merge it, then you could push it to the update side and then Martin could test the actual functionality (I think he might have time next week).
OK with you?

NicoKiaru:
Perfect with me!

Issue #72, opened 2020-01-18
SourceAndConverter naming


tischi:
We still very often say `Source` where it is a `SourceAndConverter`:

<img width="375" alt="image" src="https://user-images.githubusercontent.com/2157566/72665258-d3738700-3a06-11ea-9bc1-fe55f5884bbe.png">

I would be in favour of changing this to always saying `SourceAndConverter` or `Sac`, because otherwise one always needs to look into the code to see what it actually is.

However

1. `SourceAndConverter` is quite a bit longer than `Source`, and
1. @tpietzsch  in `bdv-core` also often says `Source` instead of `SourceAndConverter`

Not sure...

tischi:
@NicoKiaru

NicoKiaru:
Let's put `SourceAndConverter` everywhere. Let's do that when we have managed to merge the manual registration and the projection modes.

tpietzsch:
@tischi @NicoKiaru In bdv-core, I have to use `SourceAndConverter` class name because of backwards compatibility. I mostly use `source` for variable and parameter names, because it's shorter. `SourceAndConverter<?>` is the canonical "source" object as far as BDV is concerned. Unfortunate class naming but well... Pure (unconverted) `Source<?>` are never added to the BDV, and a `Source` always is necessarily associated with a converter.

Following that reasoning, I think in downstream code (vistools etc) it is ok to use "Source" in class names when it is actually a `SourceAndConverter`. I actually would prefer it that way. It's only worth thinking about, when you actually have a mix of `Source` and `SourceAndConverter` in the API. If that is mostly not the case, I would stick with "Source" (and try to hide the `Source` from public API as much as possible).

NicoKiaru:
Closing - we use Source in the naming even though the objects manipulated are SourceAndConverter(s)

PR #73, opened 2020-01-18
Projection mode changer

tischi:


tischi:
Hi @NicoKiaru ,

I mainly implemented methods to make the `ProjectionModeChanger` work: https://github.com/bigdataviewer/bigdataviewer-playground/blob/projectionModeChanger/src/test/src/sc/fiji/bdvpg/bdv/ProjectionModeChangerDemo.java#L14

Along the way, as discussed, several other things were touched, e.g. one interface was removed.

The naming conventions are not consistent, see here: https://github.com/bigdataviewer/bigdataviewer-playground/issues/72

I suggest, we fix this with another PR.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -136,9 +136,9 @@ else if ( projectionModes[sourceIndex].equals( Projection.PROJECTION_MODE_AVG ))
 			if ( source instanceof TransformedSource )
 				source = ( ( TransformedSource ) source ).getWrappedSource();
 
-			SourceAndConverter sac = getSourceAndConverter( sacs, source );
+			SourceAndConverter sac = SourceAndConverterServices.getSourceAndConverterService().getSourceAndConverterFromSource( source );
 
-			final String projectionMode = (String) SacServices.getSacService().getMetadata( sac, PROJECTION_MODE );
+			final String projectionMode = (String) SourceAndConverterServices.getSourceAndConverterService().getMetadata( sac, PROJECTION_MODE );
```
tischi, line 137: @NicoKiaru 
The `AccumulateProjectorFactory` in `bdv` work on a list of `Source` and not `SourceAndConverter`, which is quite unfortunate for us, because I need to get the metadata (`projectionMode`) for each `Source`. I thus  added a function to our `SourceAndConverterService` that returns the `SourceAndConverter` for a `Source`. I am now very confused, because in our framework (e.g. due to `SourceAndConverterDuplicator`) there could be multiple `SourceAndConverter` for the same `Source` java object, in principle with different metadata. Currently the function I added (i.e., `getSourceAndConverterFromSource`) simply returns the first matching `SourceAndConverter`.

I am not sure how to handle this properly, i.e. how to figure out to which `SourceAndConverter` the `Source` belongs.

Do you have some ideas? 

Question that come to my mind is: 

1. Do we really need the `SourceAndConverterDuplicator` or can we avoid having this? I checked for its usages and it is only used by the `SourcesDuplicatorCommand`. As the code does not need it, do your users need it or could you provide similar functionality differently?
1. If we really need the `SourceAndConverterDuplicator`, could we wrap the `Source` that we put into the new `SourceAndConverter` such that it becomes a different and unique java object? E.g., make a `DuplicatedSource( source )`?
1. What are the other places where we wrap a `Source` into a `SourceAndConverter`? Could we also there wrap the `Source` into a unique object? 
1. Could/Should we enforce that the metadata for a `Source` is unique? That is, we would not allow different `SourceAndConverter` for the same `Source` to have different metadata?

If you do not know a solution I would ask on `gitter`.

src/main/java/sc/fiji/bdvpg/services/ISourceAndConverterService.java
```diff
@@ -96,6 +96,6 @@
     /**
      * Finds the corresponding registered sac for a source.
      */
-    SourceAndConverter getSourceAndConverter( Source source );
+    SourceAndConverter getSourceAndConverterFromSource( Source source );
```
NicoKiaru: Is this function new ? I don't remember where it comes from, but since you can have multiple SourceAndConverter for an identical Source, we should remove it and keep only the function `getSourceAndConvertersOfSource`.

NicoKiaru:
Should I review it ? Is it ready ?

tischi:
> Should I review it ? Is it ready ?

Yes, please. I think we could merge it.

NicoKiaru:
OK, I start the review now

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -80,16 +79,14 @@ protected void accumulate(
 				continue;
 			}
 
-			String projectionMode = getProjectionMode( sourceIndex, sourceAndConverterToMetadata );
-
-			if ( projectionMode.equals( "Sum" ) )
+			if ( projectionModes[sourceIndex].equals( Projection.PROJECTION_MODE_SUM ) )
```
NicoKiaru: It's a function which is supposed to be fast, right ? I'd change the ".equals(string)" comparison with an int:
`projectionModes[sourceIndex] == PROJECTION_MODE_SUM` where `PROJECTION_MODE_SUM` is a `static final int`

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -80,16 +79,14 @@ protected void accumulate(
...
 			{
 				aAccu += a;
 				rAccu += r;
 				gAccu += g;
 				bAccu += b;
 			}
-			else if ( projectionMode.equals( "Avg" ))
+			else if ( projectionModes[sourceIndex].equals( Projection.PROJECTION_MODE_AVG ))
```
NicoKiaru: same as previous

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -127,26 +124,38 @@ else if ( projectionMode.equals( "Avg" ))
...
-				final Set< String > metadata = sourceAndConverterToMetadata.get( sac ).keySet();
-				if ( metadata.contains( "ProjectionMode" ) )
-				{
-					projectionMode = (String) sourceAndConverterToMetadata.get( sac ).get( "ProjectionMode" );
-				}
-			}
+			if ( source instanceof TransformedSource )
+				source = ( ( TransformedSource ) source ).getWrappedSource();
+
+			SourceAndConverter sac = SourceAndConverterServices.getSourceAndConverterService().getSourceAndConverterFromSource( source );
```
NicoKiaru: This could fail because a `Source` can have different converters and thus exists in different flavors. One Source -> 2 SourceAndConverters, one with a gray LUT and one with a completely different LUT, for instance.

We have to think about that, but that's a major issue. I had to deal with this in the remove(SourceAndConverter, BdvHandle) function from the display service, you should have a look at it and then let's rediscuss this. Maybe there's no better way...

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
```diff
@@ -127,26 +124,38 @@ else if ( projectionMode.equals( "Avg" ))
...
 		{
-			if ( sac.getSpimSource().equals( source ) )
-			{
-				final Set< String > metadata = sourceAndConverterToMetadata.get( sac ).keySet();
-				if ( metadata.contains( "ProjectionMode" ) )
-				{
-					projectionMode = (String) sourceAndConverterToMetadata.get( sac ).get( "ProjectionMode" );
-				}
-			}
+			if ( source instanceof TransformedSource )
```
NicoKiaru: This, I don't like as well. What if you want different projectors for the Source and for the transformed Source ? Currently, no Source is wrapped as a TransformedSource in our design. So I would say it's not necessary. Have you tried to remove it? Does it change anything ?

src/main/java/sc/fiji/bdvpg/bdv/projector/Projection.java
```diff
@@ -1,8 +1,13 @@
...
 
-public class ProjectionTypes
+public class Projection
 {
+	public static final String PROJECTION_MODE = "Projection Mode";
+	public static final String PROJECTION_MODE_SUM = "Sum";
+	public static final String PROJECTION_MODE_AVG = "Avg";
+
 	public static final String MIXED_PROJECTOR = "Mixed Projector";
 	public static final String SUM_PROJECTOR = "Sum Projector";
```
NicoKiaru: cf comment above, maybe change this to int to fasten the computation

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
```diff
@@ -428,56 +429,69 @@ public BdvHandle getB() {
...
      */
     class BdvHandleRef {
-        BdvHandle bdvh;
+        BdvHandle bdv;
         int indexInBdv;
 
-        public BdvHandleRef(BdvHandle bdvh, int idx) {
-            this.bdvh = bdvh;
+        public BdvHandleRef( BdvHandle bdv, int idx) {
+            this.bdv = bdv;
```
NicoKiaru: Why this change ? I thought you liked `bdvHandle` so either `bdvHandle` or `bdvh` but not `bdv` only.

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterService.java
```diff
@@ -226,6 +240,27 @@ public void registerPopupSourcesAction(Consumer<SourceAndConverter[]> action, St
...
+        if ( source instanceof TransformedSource )
+            source = ( ( TransformedSource ) source ).getWrappedSource();
+
+
+        for ( SourceAndConverter sac : sacs )
+            if ( sac.getSpimSource().equals( source ) )
+                return sac;
+
+        return null;
+    }
```
NicoKiaru: This part I have major concerns. First part :
* a Source can have muliply linked `SourceAndConverter`
* a TransformedSource is different from a Source, so we should be able to set different projectors and properties if we deal with a Source or one of its transformations

src/main/java/sc/fiji/bdvpg/services/ISourceAndConverterService.java
NicoKiaru: cf COmment, this should return a List<SourceAndConverter>

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
```diff
@@ -166,37 +167,37 @@ public void makeInvisible(SourceAndConverter sac) {
...
 
-        bdvh.getViewerPanel().addSource(sac);
-        bdvh.getSetupAssignments().addSetup(getConverterSetup(sac));
+        bdvHandle.getViewerPanel().addSource(sac);
+        bdvHandle.getSetupAssignments().addSetup(getConverterSetup(sac));
 
         // Stores where the sourceandconverter is displayed (BdvHandle and index)
-        BdvHandleRef bhr = new BdvHandleRef(bdvh, bdvh.getViewerPanel().getState().numSources());
-        if (!( sacToBdvHandleRefs.containsKey(sac))) {
+        BdvHandleRef bhr = new BdvHandleRef(bdvHandle, bdvHandle.getViewerPanel().getState().numSources());
```
NicoKiaru: This sort of refactoring is a bit annoying for the reviewer. Let's try to refactor less, and in the master. It will be more straightforward for reviews to see what are the change that actually matter

NicoKiaru: This sort of refactoring is a bit annoying for the reviewer. Let's try to refactor less, and in the master. It will be more straightforward for reviews to see what are the change that actually matter

NicoKiaru (changes requested):
We should address the problem between linking Source and SourceAndConverter

NicoKiaru:
The link between `Source`and `SourceAndConverter` is done in the method `createProjector` of the class `MultiResolutionRenderer`.

With:
* the function: `bdvHandle.getViewerPanel().getState().getVisibleSourceIndices()` 
* a reference to BdvHandle being used by the factory
* an access to `BdvHandleReferences` of the display service

We should be able to reconstitute the sequence that led to the projector factory call, and retrieve from which `SourceAndConverter` a `Source` comes from.

Another option would is to get the first SourceAndConverter that comes (as it is currently but it will fail sometimes). But we could make this assumption for now and see if the new BigDataViewer state in bigdataviewer/bigdataviewer-core#68 allows us to find more easily the link between Source and SourceAndConverter

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
tischi: True! Let's try to avoid this.

tischi: True! I'll try to avoid this!

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterService.java
tischi: Yes, I had the same problem. I wrote a some text about it below, which you probably did not see. I paste it here again:

---

The AccumulateProjectorFactory in bdv work on a list of Source and not SourceAndConverter, which is quite unfortunate for us, because I need to get the metadata (projectionMode) for each Source. I thus added a function to our SourceAndConverterService that returns the SourceAndConverter for a Source. I am now very confused, because in our framework (e.g. due to SourceAndConverterDuplicator) there could be multiple SourceAndConverter for the same Source java object, in principle with different metadata. Currently the function I added (i.e., getSourceAndConverterFromSource) simply returns the first matching SourceAndConverter.

I am not sure how to handle this properly, i.e. how to figure out to which SourceAndConverter the Source belongs.

Do you have some ideas?

Question that come to my mind is:

Do we really need the SourceAndConverterDuplicator or can we avoid having this? I checked for its usages and it is only used by the SourcesDuplicatorCommand. As the code does not need it, do your users need it or could you provide similar functionality differently?
If we really need the SourceAndConverterDuplicator, could we wrap the Source that we put into the new SourceAndConverter such that it becomes a different and unique java object? E.g., make a DuplicatedSource( source )?
What are the other places where we wrap a Source into a SourceAndConverter? Could we also there wrap the Source into a unique object?
Could/Should we enforce that the metadata for a Source is unique? That is, we would not allow different SourceAndConverter for the same Source to have different metadata?
If you do not know a solution I would ask on gitter.

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
tischi: I think I asked in an issue whether it would be ok to change `bdvHandle` => `bdv`, and I think you said yes, but we could also stick with `bdvHandle`.

src/main/java/sc/fiji/bdvpg/bdv/projector/Projection.java
tischi: Ok. I could do this, but long term I think an enum would be the cleanest. 
The only reason I was choosing String constants was because this is the only thing that works in `Command` choices. Thus, using `int`s not would make it more complicated because we have to implement the constants twice.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: I agree, I also don't like it but I was not sure how this is going to be in the end with your code.
I can remove it!

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: Agreed! See conversations below.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: I would leave it as a string for now unless we feel performance issues.
See my answer below.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: see below.

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: Hi @tpietzsch,
We want to implement `SourceAndConverter`-dependent projection modes. For this to work, we currently look within the `AccumulateProjector` in a global map which projection mode is associated with which `SourceAndConverter`. Now, unfortunately the `AccumulateProjector` only knows about a `List< Source >` and not a `List< SourceAndConverter >`. This is a challenge for us, because one java `Source` object could correspond to multiple `SourceAndConverter`. Thus we do not know how to properly do the mapping. Do you have some advise?


@NicoKiaru

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
NicoKiaru: ok, maybe I misunderstood, I think it's good to keep bdvHandle then. It's more consistent with what we have everywhere else.

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterService.java
NicoKiaru: Here's how I dealt with this issue in the remove from display :

https://github.com/bigdataviewer/bigdataviewer-playground/blob/c4f9b34607811e23180b01a8d6eee279cedff522/src/main/java/sc/fiji/bdvpg/scijava/services/BdvSourceAndConverterDisplayService.java#L249-L256

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tischi: @NicoKiaru

As discussed, I managed to inject a `bdvHandle` into the `AccumulateMixedProjectorARGB` (https://github.com/bigdataviewer/bigdataviewer-playground/pull/73/commits/df97e5db522db2c3c087546e09255dfee2a7e890)

There is now a function
`getProjectionModes( BdvHandle bdvHandle, ArrayList< Source< ? > > sources )`
from which you could continue working on, as discussed.

Thanks!

tpietzsch: @tischi @NicoKiaru I'll change it, so that `createAccumulateProjector()` has a
`List<SourceAndConverter<?>> sources` parameter instead of `List<Source<?>>`.

NicoKiaru: Thanks @tpietzsch !
In the new Improved UI branch or in the old one ?

tischi: >  so that createAccumulateProjector() has a
List<SourceAndConverter<?>> sources parameter instead of List<Source<?>>.

awesome, that was in fact my dream :-)

tpietzsch: @NicoKiaru I'll do it in the new one, probably later today

NicoKiaru: Thanks for your input!

src/main/java/sc/fiji/bdvpg/bdv/projector/Projection.java
NicoKiaru: Ok, I'll investigate the performance and report back

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
NicoKiaru: Resolved in old version. Need to check on the branch for the new ui how to solve it in more simple manner

NicoKiaru (approved):
Looks all ok

tischi:
Great work!
Thanks a lot!

src/main/java/sc/fiji/bdvpg/bdv/projector/AccumulateMixedProjectorARGB.java
tpietzsch: @NicoKiaru @tischi the new version would be in this branch
https://github.com/bigdataviewer/bigdataviewer-core/tree/projector-factory
(on top of the new ui branch)
Do you think this is useful?

tischi: @tpietzsch That would be wonderful! 
Just to make sure: can we rely on the `SourceAndConverter` objects in the list being identical to the ones that we add using `bdvHandle.getViewerPanel().addSource( sac )`? Or could there be some wrapping happening inside `bdv`?

tpietzsch: Yes, you can rely on that they are `.equals()`.

At the moment, they are even the same object `==`, but I would that might change at some point, so better to use `.equals()`.

tischi: Then I think it would be perfect.
@NicoKiaru agree?

NicoKiaru: of course, looks perfect. We have several workarounds at the moment to make this work for the old version, but that would be reeally nice to remove them.

tpietzsch: Ok, perfect, then I'll just merge it to the "new ui" branch and then that should be merged to master pretty soon

PR #75, opened 2020-01-19
Add context menu behaviour

tischi:
builds upon: https://github.com/bigdataviewer/bigdataviewer-playground/pull/73

@NicoKiaru 
In this branch, I'll be working on a context menu for a right-click within a bdv, showing a menu with actions to perform on any source under the mouse location.


tischi:
@NicoKiaru FYI: Above it also shows your comments to the other PR. This is a bit confusing. I do not know how to change this.

tischi:
Next steps:

- [x] Only apply commands to Sources under the mouse cursor
- [ ] Implement something that makes is possible to select for which Sources under the mouse cursor the actions should be applied (...could be done later otherwise this PR becomes too lagre)

tischi:
@NicoKiaru 
This could also be merged, if you agree.

NicoKiaru:
@tischi could you try to merge the master into this branch at this point ? Then I could have a look at it and merge it into master

tischi:
@NicoKiaru 

1. I merged the latest master
1. To try the functionality, you can run `SpimDataDisplayDemo` and then press `c` on top of a source in the bdv window.
1. In `SourceAndConverterPopupMenu` I put some static methods that we could also use in `BdvSourceServiceUI` to avoid code duplication. I did not do this yet to not make too many changes.

src/main/java/SourceA.java
```diff
@@ -0,0 +1,3 @@
+public class SourceA
```
NicoKiaru: Shouldn't this be removed ?

src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
```diff
@@ -53,17 +55,32 @@ private void createEmptyBdv()
...
 
-		bdvHandle = bss.getBdvHandle();
+		bdv = bss.getBdvHandle();
 
-		if ( interpolate ) bdvHandle.getViewerPanel().setInterpolation( Interpolation.NLINEAR );
+		if ( interpolate ) bdv.getViewerPanel().setInterpolation( Interpolation.NLINEAR );
 
 		bss.removeFromBdv();
+
+		addBehaviours();
```
NicoKiaru, line 64: This should not go here : this sort of action has to be configurable, not present by default in bdv. Maybe somebody wants to bind right click with something else, like in the case of labeleditor.

src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
```diff
@@ -53,17 +55,32 @@ private void createEmptyBdv()
...
+
+	private void addSourceAndConverterContextMenuBehaviour()
+	{
+		final ClickBehaviourInstaller installer = new ClickBehaviourInstaller( bdv, new SourceAndConverterContextMenuClickBehaviour( bdv ) );
+
+		installer.install( "Sources context menu - C", "C" );
+		installer.install( "Sources context menu - Right mouse button", "button3" );
 	}
 
 	public BdvHandle get()
```
NicoKiaru, line 80: We should remove the functional interface here. I think it's confusing, and anyway we do not want to supply thousands of Bdv windows

src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
NicoKiaru: Action the addclick behaviour is broken in newBdvCore because we use BigDataViewer instead of BdvHandle. Let's try to keep things separate to fix them more easily.

src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
```diff
@@ -293,5 +296,14 @@ public static ARGBType getSourceColor(BdvHandle bdvHandle, int sourceId )
         return new double[]{ displayRangeMin, displayRangeMax };
     }
 
+    public static RealPoint getPhysicalMouseCoordinates( BdvHandle bdv )
+    {
+        final RealPoint posInBdvInMicrometer = new RealPoint( 3 );
```
NicoKiaru: Position has no reason to be in micrometer. I use the position in millimeters in some cases

src/main/java/sc/fiji/bdvpg/behaviour/SourceAndConverterContextMenuClickBehaviour.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@Override
+	public void click( int x, int y )
+	{
+		showPopupMenu( bdv, x, y );
+	}
+
+	private static void showPopupMenu( BdvHandle bdv, int x, int y )
+	{
+		final List< SourceAndConverter > sacs = SourceAndConverterUtils.getSacsAtMousePosition( bdv );
```
NicoKiaru: Is it the correct location for such a function ? I'd keep it inside this class maybe. But I'm not sure.

src/main/java/sc/fiji/bdvpg/behaviour/SourceAndConverterContextMenuClickBehaviour.java
```diff
@@ -0,0 +1,40 @@
...
+
+		if ( sacs.size() == 0 )
+			return;
+
+		SourceAndConverter[] sacArray = new SourceAndConverter[sacs.size()];
+		sacArray = sacs.toArray(sacArray);
+
+		final SourceAndConverterPopupMenu popupMenu = new SourceAndConverterPopupMenu( sacArray );
```
NicoKiaru, line 47: Don't you want the popup menu to be configurable ?

src/main/java/sc/fiji/bdvpg/behaviour/SourceAndConverterContextMenuClickBehaviour.java
NicoKiaru: Or maybe change the argument to coordinate values x,y,z of the absolute coordinate instead. Let me know what you think

src/main/java/sc/fiji/bdvpg/behaviour/SourceAndConverterContextMenuClickBehaviour.java
NicoKiaru: I think we need to think differently with these actions on SourceAndConverters.

We need to modularize this and maybe access bdv actions only through the action Map<String, Consumer<SourceAndConverter[]>>

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
```diff
@@ -625,4 +625,13 @@ void removeSourceViaReflection(BdvHandle bdvh, int index) {
         }
     }
 
+    public List< SourceAndConverter > getSourceAndConverters( BdvHandle bdv )
+    {
+        return bdvHandleToSacs.get( bdv );
+    }
+
+    public CommandService getCommandService()
```
NicoKiaru: I'm not sure it's a good idea to ask for a service within another service. Why do need this ?

Getting the action map Map<String, Consumer<SourceAndConverter[]>> should be sufficient

src/main/java/sc/fiji/bdvpg/scijava/services/ui/BdvSourceServiceUI.java
```diff
@@ -485,10 +485,10 @@ void visitAllNodesAndDelete(TreeNode node, SourceAndConverter sac) {
      * @param action
      * @param actionName
      */
-    public void addPopupAction(Consumer<SourceAndConverter[]> action, String actionName) {
+    public void addPopupAction( Consumer< SourceAndConverter[] > action, String actionName ) {
```
NicoKiaru: That's what we need to modify as well

src/main/java/sc/fiji/bdvpg/services/ISourceAndConverterService.java
```diff
@@ -98,4 +99,6 @@
      */
     List<SourceAndConverter> getSourceAndConvertersFromSource( Source source );
 
+    BdvSourceServiceUI getUI();
```
NicoKiaru: Let's try to keep the UI  separated and just keep actions

src/main/java/sc/fiji/bdvpg/source/SourcePhysicalIntervalDeterminer.java
```diff
@@ -0,0 +1,27 @@
...
+import net.imglib2.util.Intervals;
+
+public class SourcePhysicalIntervalDeterminer
+{
+	private final Interval interval;
+
+	public SourcePhysicalIntervalDeterminer( Source< ? > source )
+	{
+		final AffineTransform3D sourceTransform = new AffineTransform3D();
+		source.getSourceTransform( 0, 0, sourceTransform );
```
NicoKiaru: Here we assert a source at timepoint 0 exists

src/main/java/sc/fiji/bdvpg/sourceandconverter/SourceAndConverterUtils.java
```diff
@@ -362,4 +369,16 @@ private static Converter createConverterARGBType(Source source) {
         return converter;
     }
 
+    public static List< SourceAndConverter > getSacsAtMousePosition( BdvHandle bdv )
```
NicoKiaru: It would be nice to add some comments here

NicoKiaru (commented):
We need to separate more clearly actions vs the ui of SourceAndConverter and Command Service. Otherwise this will very annoying to maintain in the long term.

I suggest that I will rewrite the SourceAndConverter service such as there will be an action map from String to Consumer<SourceAndConverter[]>. This will be the way to comminucate, and build ui based on this map.

Also, could you test on whether warped source are supported by this new popup window ? And how does it behave with source with no  bounds (a generative 3d wave). Will it work ? Not that it's necessary for it to work in all cases, but it would be nice to mention the potential limitations, if any.

src/main/java/sc/fiji/bdvpg/services/ISourceAndConverterService.java
NicoKiaru: https://github.com/bigdataviewer/bigdataviewer-playground/pull/75/commits/99f88c9861779dab544fc05b71ac09135e864301 revert this

src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java
NicoKiaru: Solved in the new structure

NicoKiaru:
So I tested it and it's really nice to access the functions from the bdv window with a right click.

There are a few structure design where we need to work, but that will be doable. Now there's a really strong limitation which I like to address now : warped sources are not handled correctly.

Do you know if there's a way to handle these warped source correctly  @tischi ?

Another potential enhancement : if we find a way to highlight which sources are selected when we right click that would be nice also.

tischi:
> Do you know if there's a way to handle these warped source correctly @tischi ?

Could you elaborate what you mean by `warped`?

> Another potential enhancement : if we find a way to highlight which sources are selected when we right click that would be nice also.

Not sure that is what you mean, but I even thought of a selection menu of all the sources that are below the mouse cursor (like the one that you have for your big warp launcher where you select the moving/target sources). Because, if there are many sources on top of each other one would not always like to apply the action to all of them. Maybe after selecting the action, there could be - only if there are multiple - another UI popping up, allowing one to only select a subset of the sources.

NicoKiaru:
> Could you elaborate what you mean by `warped`?

To see the issue:
* Launch BigWarp and do a basic warping. 
* The warped sources are instances of [`WarpedSource`](https://github.com/saalfeldlab/bigwarp/blob/9f1802bc28201162d4c5f65d98c654e2b09d3c4e/src/main/java/bdv/img/WarpedSource.java). You can use the inspect function.
* show one of these source in a new bigdataviewer window. If you right click at different locations, you will notice that the source is not detected correctly on all places. I can make a screencast if you wish, let me know.

I believe with the current design, only affinetransforms are handled correctly.

src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java
NicoKiaru: Minor, let's keep it like that

Issue #82, opened 2020-02-05
capture view


tischi:
@NicoKiaru I need to implement a new view capture that simply returns a RGB image.
You once mentioned that you also have code for capturing the current view. Could you point me to it to have a look?

NicoKiaru:
I don't know if that helps, but here's my implementation:

https://github.com/BIOP/bigdataviewer_scijava/blob/master/src/main/java/ch/epfl/biop/bdv/scijava/command/export/BdvSourcesBdvViewToImagePlus.java

It outputs RGB image, but only if the source is RGB.

tischi:
Looks good......  
Do you have a feeling for whether this is significantly different from what we currently have?

https://github.com/bigdataviewer/bigdataviewer-playground/blob/be40dd31bf4a2ead6a02894f57aa975d20188d92/src/main/java/sc/fiji/bdvpg/bdv/ScreenShotMaker.java#L77

Just glancing over both codes I had the feeling it might be more or less the same?

NicoKiaru:
I may have forgotten things, but here's how it compares

Supported in my implementation and not in screenshotmaker (to my knowledge):
- All RealType Source (FloatType to float for instance) + ARGBSource to RGB ImagePlus
- Export of ImageJ1 ImagePlus Calibration (XYZ spacings and offsets)
- Export of z stack (with different z spacing)

Supported in ScreenShotMaker and not in mine:
- Automatic choice of mimap level
- Segmentation ? Glasbey image

Difference in implementation (not good or bad, may have performance difference depending on use cases):
- Virtual stack vs non virtual stack
- My impl : parallelization made on sources vs ScreenShotMaker : parallelization wihin one image with grid access
- Compulsory explicit choice of xy voxel size in mine
- Compulsory explicit choice of source in mine

tischi:
code is still horrible, but I think I implemented a good functionality:
https://github.com/embl-cba/fiji-plugin-platyBrowser/issues/47

martinschorb:
Is this functionality available in the current version?
Would typing C do it? 
Instead of opening a new ImageJ window, could I create a stack/movie by pressing c multiple times while browsing/zooming/rotating through the volume?

tischi:
to make a movie I would probably just use some screen casting tools like QuickTime Player.

tischi:
regarding typing C: I still have to create a PR to add the functionality.

martinschorb:
[...]

PR #91, opened 2020-02-28
Addresses #82 - Add raw and rgb screenshot functionality

tischi:


src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
```diff
@@ -298,4 +297,11 @@ public static ARGBType getSourceColor(BdvHandle bdvHandle, int sourceId )
         return new double[]{ displayRangeMin, displayRangeMax };
     }
 
+	public static Converter< ?, ARGBType > getConverter( Bdv bdv, int sourceIndex )
```
NicoKiaru, line 321: This is supposed to disappear (indexing logic). Can't you use directly SourceAndConverter instead ? There's a way to retrieve all SourceAndConverters from a Bdv window with the SourceAndConverterDisplayService (https://github.com/bigdataviewer/bigdataviewer-playground/blob/bf0e03687828d70d23822287937d719911d2c4ff/src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java#L584), and also there's a function to know if it's visible (https://github.com/bigdataviewer/bigdataviewer-playground/blob/bf0e03687828d70d23822287937d719911d2c4ff/src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterBdvDisplayService.java#L160)

NicoKiaru (commented):
I don't grasp everything that' happening here. WHen you have time, it would be nice to have comments in ScreenSHotMaker

NicoKiaru:
You can resolve the conflicts and then I can merge it

src/main/java/sc/fiji/bdvpg/bdv/BdvUtils.java
tischi: done. thanks for the suggestions!

tischi:
@NicoKiaru I would like to add the ScreenShot functionality somewhere where it is easy to access.

1. Should I add it to the PopupMenu? I do not manage, because I think only actions that consume `sac`s are welcome there.

2. Should I add it as a keyboard shortcut to the `BdvCreator`? Which shortcut would you suggest? `C` is already taken now... maybe `shift S` for **S**creenshot? `S` is already defined within Bdv.

tischi:
@NicoKiaru Can you maybe help?
When running the `ScreenshotDemo` it only find one of the two sources visible. I do not understand why and have trouble debugging your code for that. 
Could you help me by debugging here: https://github.com/bigdataviewer/bigdataviewer-playground/pull/91/commits/23453569e0e3a59c0d3625baad9c255f7ae624d5 ?
As said, I think it should evaluate **both** `sac`s to `true`

NicoKiaru:
> 
> 
> @NicoKiaru Can you maybe help?
> When running the `ScreenshotDemo` it only find one of the two sources visible. I do not understand why and have trouble debugging your code for that.
> Could you help me by debugging here: [2345356](https://github.com/bigdataviewer/bigdataviewer-playground/commit/23453569e0e3a59c0d3625baad9c255f7ae624d5) ?
> As said, I think it should evaluate **both** `sac`s to `true`

So I hope the isvisible issue should be resolved

NicoKiaru:
Ok, let's merge this, @tischi  ? 

An enhancement would be to use the same projector as the bdv window. because now it ignores the occluding part, for instance.

I could open an enhancement issue or if you feel this is simple you can do it directly before merging ?

tischi:
@NicoKiaru 
I had a look; some work, but doable before merging. However we would need to be able to know which AccumulateProjectionFactory is used for a given bdvHandle. Do you have an idea how to do this best?

tischi:
@NicoKiaru 
Ok, works now. Required more changes than expected. If you run the ScreenShotDemo you could try it out. 

Question: How do we make the Screenshot accessible to the user?  **Should we add a Keyboard shortcut for now**? And maybe later add it to the context menu?

NicoKiaru:
Cool! I'll try it.

If you want to add the shortcut to BdvCreator for now that's fine.

We'll have to modularize this later when needed.

tischi:
To add this to `BdvCreator` I need the `CommandService` within `BdvCreator`, because I need to run `ScreenShotMakerCommand`. How should I do this? Can I get it from somewhere?

NicoKiaru:
Can't you run the action directly in 
https://github.com/bigdataviewer/bigdataviewer-playground/blob/64bee42b70f731190e5c628887aefa401436b179/src/main/java/sc/fiji/bdvpg/bdv/BdvCreator.java#L72

and skipping the commandservice ?

tischi:
But there should be the Command UI popping up, for this I need `ij.command().run( ScreenShotMakerCommand.class )` or not?

tischi:
The `SourceAndConverterBdvDisplayService` has a `CommandService`.
We could add a `get` method. It does not feel very logical though...

Other option: we add a `CommandService` as another service to `SourceAndConverterServices`, like `SourceAndConverterServices.getCommandService()`.

NicoKiaru:
you can have a look at the former commit

* I had to tweak the 'actions' registering in the service, because they are supposed to get sourceandconverters as inputs. 

I think we could avoid this and make screenshot a bit more elegant:
* If it's launched with the key from a bdv window : send all visible sources by default
* If it's launched from a FIJI menu : let the user select everything or just a subset of the sources

This last point needs a bit of tweaking if we want the user to have displayed only the sources present in the bdv window

We can also let it like that

tischi:
@NicoKiaru did you look into the bug yet? 
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at sc.fiji.bdvpg.bdv.BdvCreator.lambda$addSourceAndConverterContextMenuBehaviour$0(BdvCreator.java:84)
	at org.scijava.ui.behaviour.MouseAndKeyHandler.handleKeyPressed(MouseAndKeyHandler.java:596)
	at org.scijava.ui.behaviour.MouseAndKeyHandler.keyPressed(MouseAndKeyHandler.java:504)
```

NicoKiaru:
DId you retry ? Do a 'clean'

I cannot reproduce the bug ...

tischi:
Works now also for me. Merging.

PR #92, opened 2020-02-28
Fixes #87 Add exclusive projection mode

### LINKED ISSUE #87: additional Projection mode
Hi,

when dealing with multi-scale datasets (large differences in magnification) it would make a lot of sense to have an additional projection mode that would display sources according to their resolution/voxel size.

Ideal scenario would be: 

- display these "Top-layer" sources on top of the "Average" sources but before applying the "Sum" sources.
- or make it optional whether they are shown in between "Avg" and "Sum" or on top of everything.

Cheers,
Martin

tischi:


NicoKiaru:
Ok, I don't understand what this is supposed to do. Can you add a comment in the PR or better - in the class, so that we can understand what are the projection modes and use cases

tischi:
@NicoKiaru Thank you for the suggestion. I added some documentation here: https://github.com/bigdataviewer/bigdataviewer-playground/pull/92/commits/b5fdec9d4297ab65e7313729809e3587df893d76

NicoKiaru:
Let's find a moment to discuss all of this at Neubias ?

I'm lagging behind

tischi:
> Let's find a moment to discuss all of this at Neubias ?

Yes! Are you coming?

src/main/java/sc/fiji/bdvpg/bdv/projector/Projection.java
```diff
@@ -1,13 +1,35 @@
...
+ *
+ * PROJECTION_MODE_SUM
+ * The ARGB values of the source will be added to the final ARGB to be displayed.
+ *
+ * PROJECTION_MODE_AVG
+ * The ARGB values of all sources with the Average projection mode will first be averaged
+ * before being added to the final ARGB to be displayed.
+ * This is useful for overlapping electron microscopy data sets.
+ *
+ * PROJECTION_MODE_EXCLUSIVE
```
NicoKiaru, line 16: Maybe explain what happens if we set two sources to ' exclusive'. If they overlap ? If they do not overlap ?

NicoKiaru: Ok in fact I still do not understand. There can be multiple 'exclusive' mode ? And you can have 'sum exclusive' and 'average exclusive'  ? What does that mean ? Also what I didn't get before was that it's a 'flag' added on top of the projection mode.

NicoKiaru (commented):
Honestly it would be better to avoid string comparison in a per pixel operation.

Having a number per projection and saving a bit to flag for 'exclusiveness' would be better practice.

The word 'exclusiveness' is a bir weird because several can be exclusive. Why not 'exclusive group' or 'occluding group' ?

src/main/java/sc/fiji/bdvpg/bdv/projector/Projection.java
NicoKiaru: Now I undertstand. Maybe occluding group could be a better name ?

NicoKiaru:
@tischi :
* let's keep the string comparison
* modify the name exclusive to something like 'add to occluding group' or whatever you prefer
* open an issue with 'enhancement' for the string comparison and potentially multilayer

-> then we merge.

Sounds ok ?

tischi:
- let's keep the string comparison

ok

- modify the name exclusive to something like 'add to occluding group' or whatever you prefer

done

- open an issue with 'enhancement' for the string comparison and potentially multilayer

done

PR #96, opened 2020-03-05
adds metadata to displays / bdv handle in the sourceandconverterdispl…

NicoKiaru:
…ay service


NicoKiaru:
@tischi you can merge if it looks ok to you

PR #97, opened 2020-03-12
adds interpolate option in source resampling + adds source creation (…

NicoKiaru:
…3d sampler)

Hi @tischi, 

Do you think this PR is ok ? It's adding a function to create a source from another one but with a different voxel sampling.

Best, 

Nico

### DIFF
--- a/src/main/java/sc/fiji/bdvpg/sourceandconverter/importer/NewSourceAndConverterGetter.java
+++ b/src/main/java/sc/fiji/bdvpg/sourceandconverter/importer/NewSourceAndConverterGetter.java
@@ -12,6 +12,16 @@ import sc.fiji.bdvpg.sourceandconverter.SourceAndConverterUtils;
 
 import java.util.function.Supplier;
 
+/**
+ * Action which generates a new Source which samples and spans a space region
+ * defined by either :
+ * - an affine transform and a number of voxels
+ * - or a model source and voxel sizes
+ *
+ * Mipmap unsupported
+ * TimePoint 0 supported only TODO : improve timepoint support
+ */
+
 public class NewSourceAndConverterGetter implements Runnable, Supplier<SourceAndConverter> {
 
     AffineTransform3D at3D = new AffineTransform3D();
@@ -22,6 +32,15 @@ public class NewSourceAndConverterGetter implements Runnable, Supplier<SourceAnd
 
     ImgFactory imgfactory;
 
+    /**
+     * Simple constructor
+     * @param name
+     * @param at3D
+     * @param nx
+     * @param ny
+     * @param nz
+     * @param imgfactory
+     */
     public NewSourceAndConverterGetter(
             String name,
             AffineTransform3D at3D,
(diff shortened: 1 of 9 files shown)


src/main/java/sc/fiji/bdvpg/sourceandconverter/importer/NewSourceAndConverterGetter.java
```diff
@@ -0,0 +1,142 @@
...
+            ) {
+        this.nx = nx;
+        this.ny = ny;
+        this.nz = nz;
+        this.at3D = at3D;
+        this.name = name;
+        this.imgfactory = imgfactory;
+    }
+
+    public NewSourceAndConverterGetter(
```
tischi: Could you give this class a more specific name? This is very generic 😄 at least I cannot guess from the name what it is doing.... 

In addition, is it maybe more a `Creator` or a `Factory` than a `Getter`? With `Getter` I associate that I get an object that exists already...

tischi (commented):
And could you try to make whatever happens in `ResampledSource` easier to understand?
I guess `ResampledSource` is the place where the main stuff happens?
I think a way could be to split up the code in `getSource` into smaller functions and give those functions names that are very descriptive.

NicoKiaru:
Come on, it's so easy to understand, no need for comments!

ok ok I'll comment a bit ;-)

Briefly (I know it's impossible to understand anything right now) : the idea is, if you have a complex source (warped several times, affine transformed, etc...) , to be able to 'resample' it as a normal 3d image with a certain voxel size. Then it can be exported as a new dataset or imageplus.

I'll put some comments and a demo of it.

tischi:
> Briefly (I know it's impossible to understand anything right now) : the idea is, if you have a complex source (warped several times, affine transformed, etc...) , to be able to 'resample' it as a normal 3d image with a certain voxel size. Then it can be exported as a new dataset or imageplus.

Ok, does that just mean (i) to add another "resample view" on top or (ii) to make an actual RAM copy where all views/ converters are applied? If it is (ii) it may often not be possible due to limiting RAM, right?

src/main/java/sc/fiji/bdvpg/sourceandconverter/importer/NewSourceAndConverterGetter.java
NicoKiaru: Changed to `EmptySourceAndConverterCreator` in https://github.com/bigdataviewer/bigdataviewer-playground/pull/97/commits/612860d6ac98ae3016a6b4353fc0fb049afa362a

NicoKiaru:
> And could you try to make whatever happens in `ResampledSource` easier to understand?

I've added documentation everywhere and a demo `sc.fiji.bdvpg.bdv.sourceAndConverter.resample.ResamplingDemo`.

Check the demo to see how it could be useful (especially for source warped by BigWarp - I'll make a screencast one day) and I hope the documentation helps in `ResampledSource`. I do not believe it is easy to split `ResampledSource` in smaller parts.

NicoKiaru:
Hi @tischi 

I'll work on trying to solve some issue that @martinschorb found today.

Are you ok with this PR ? I'd like to merge it if possible in the release.

If you think not, I can move the functions in another repo. Just let me know.

Best, 

Nico

NicoKiaru:
Hey @tischi I will merge this PR because I need to move on and solve Martin issue. I need to have this PR because the version in the update site is too intricated with this branch. Sorry about this

tischi:
No problem. I will only have time middle/end of next week again. I have to prepare a two day online course on image analysis...........

PR #105, opened 2020-03-28
Orthoview

NicoKiaru:
Makes possible to have three synchronized windows to get orthogonal view of the sources.

Two useful commands:
* Bdv>Create Ortho Bdv Frames
* Bdv>Show Multiple sources in multiple windows : to add sources to the three windows at the same time

image

What could be great is to add a cross as an overlay at the center of the windows

@tischi if that sounds good let me know
ping @martinschorb


NicoKiaru:
Now contains cross overlay

martinschorb:
My ortho windows are not centered at 0
Also, when I open a source, I cannot find it anywhere ;-(

UPDATE:
Cool: AdjustView or Right-click->Show Source  work in the Ortho-Windows as well to get the right viewer focus!!!

martinschorb:
Oh, and is it possible to disable the left mouse drag rotation for these ortho-windows? That really screws it up...

NicoKiaru:
> 
> 
> My ortho windows are not centered at 0
> Also, when I open a source, I cannot find it anywhere ;-(
> 
> UPDATE:
> Cool: AdjustView or Right-click->Show Source work in the Ortho-Windows as well to get the right viewer focus!!!

Yes, you can also use the command `Bdv>Show Sources in Multiple Bdv Windows`, and select all three ortho views

NicoKiaru:
> 
> 
> Oh, and is it possible to disable the left mouse drag rotation for these ortho-windows? That really screws it up...

Arf, that's a modification of the standard which I do not know. I'm not sure it's a good idea to deviate from the way it's working with other bdv plugins.

But to get back right the orientation, is the shorcut `Shift +x, y, or z` not sufficient ?

PR #127, opened 2020-05-11
Bdvhandlenodes

NicoKiaru:
This branch should allow to add custom nodes in the tree of the `SourceAndConverter` service.

Challenge:
* make these filter customizable
* handle synchronization and avoiding building the tree each time there's an event
* avoid Swing synchronization bugs

Optional:
* support custom rendering
* support adequate context menu = context is the node / nodes clicked


NicoKiaru:
At this point:
- fixed synchronization 
- refactored filtering mechanism
- nodes added which are showing sources displayed in the BdvHandles

PR #157, opened 2020-11-02
Add UnmodifiableConverterSetup

tischi:
Enabling adding Converters that we do not actively support yet.

### DIFF
--- a/src/main/java/bdv/util/UnmodifiableConverterSetup.java
+++ b/src/main/java/bdv/util/UnmodifiableConverterSetup.java
@@ -1,0 +1,103 @@
+/*-
+ * #%L
+ * BigDataViewer-Playground
+ * %%
+ * Copyright (C) 2019 - 2020 Nicolas Chiaruttini, EPFL - Robert Haase, MPI CBG - Christian Tischer, EMBL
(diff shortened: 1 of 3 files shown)


NicoKiaru:
What's the use case ?

Also a message should still be displayed with the class of Converter in order to be able to implement it appropriately.

NicoKiaru:
Also this does not concerns `Converters` but `ConverterSetups`. So I don't see how this improves our support of Converters ?

tischi:
> What's the use case ?

If you look at the `getConverter()` method of `SourceAndConverter` anything that implements `Converter< T, ARGBType >` is valid. However, we only supported few specific implementations (`ColorConverter` and `RealLUTConverter`). For my project, I have sources with another converter and would like to show them (I will check though whether I can still manage to implement `ColorConverter`). Anyway, I think it is good to also support other converters than `ColorConverter` and `RealLUTConverter`.

> Also a message should still be displayed with the class of Converter in order to be able to implement it appropriately.

Done.

> Also this does not concerns Converters but ConverterSetups. So I don't see how this improves our support of Converters ?

See answer above.

PR #159, opened 2020-11-03
Improve ViewerTransformChanger and add AdaptedViewCenterTransformCreator

tischi:
I added more functionality to the navigate package.

You can see it in practice by running the amended `ViewTransformSetAndLogDemo`.

### DIFF
--- a/src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
+++ b/src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
@@ -1,0 +1,67 @@
+package sc.fiji.bdvpg.bdv.navigate;
+
+import bdv.util.Bdv;
+import bdv.util.BdvHandle;
+import bdv.viewer.animate.SimilarityTransformAnimator;
+import net.imglib2.realtransform.AffineTransform3D;
+
+public class AdaptedViewCenterTransformCreator
+{
+	private final BdvHandle bdvHandle;
+	private final double[] xyz;
+	private final int t;
+
+	/**
+	 * Based on the current ViewerTransform, creates a new viewer transform with a new center position.
+	 *
+	 * @param bdvHandle
+	 * @param xyz
+	 * 			target coordinates in physical units
+	 * @param t
+	 * 			time point
+	 */
+	public AdaptedViewCenterTransformCreator( BdvHandle bdvHandle, double[] xyz, int t )
+	{
+		this.bdvHandle = bdvHandle;
+		this.xyz = xyz;
+		this.t = t;
+	}
+
+	private AffineTransform3D createAdaptedTransform()
+	{
+		bdvHandle.getViewerPanel().setTimepoint( t );
+
+		final AffineTransform3D currentViewerTransform = new AffineTransform3D();
+		bdvHandle.getViewerPanel().state().getViewerTransform( currentViewerTransform );
+
+		AffineTransform3D adaptedViewerTransform = currentViewerTransform.copy();
+
+		// ViewerTransform notes:
+		// - applyInverse: coordinates in viewer => coordinates in image
+		// - apply: coordinates in image => coordinates in viewer
+
+		final double[] targetPositionInViewerInPixels = new double[ 3 ];
+		currentViewerTransform.apply( xyz, targetPositionInViewerInPixels );
+
+		for ( int d = 0; d < 3; d++ )
+		{
+			targetPositionInViewerInPixels[ d ] *= -1;
+		}
+
+		adaptedViewerTransform.translate( targetPositionInViewerInPixels );
+
+		final double[] windowCentreInViewerInPixels = new double[ 3 ];
+		windowCentreInViewerInPixels[ 0 ] = bdvHandle.getViewerPanel().getDisplay().getWidth() / 2.0;
+		windowCentreInViewerInPixels[ 1 ] = bdvHandle.getViewerPanel().getDisplay().getHeight() / 2.0;
+
+		adaptedViewerTransform.translate( windowCentreInViewerInPixels );
+
+		return adaptedViewerTransform;
+	}
+
+	public AffineTransform3D getAdaptedCenterTransform()
+	{
+		AffineTransform3D adaptedTransform = createAdaptedTransform();
+		return adaptedTransform;
+	}
+}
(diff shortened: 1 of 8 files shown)


src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
```diff
@@ -0,0 +1,65 @@
...
+import net.imglib2.realtransform.AffineTransform3D;
+
+public class AdaptedViewCenterTransformCreator
+{
+	private final BdvHandle bdvHandle;
+	private final double[] xyz;
+	private final int t;
+
+	/**
+	 * Based on the current ViewerTransform, creates a new viewer transform with a new center position.
```
NicoKiaru, line 13: Maybe move this to the header and put the documentation like the other actions.

Also I think we were trying to implement Runnable.

src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
```diff
@@ -0,0 +1,65 @@
...
+		final double[] windowCentreInViewerInPixels = new double[ 3 ];
+		windowCentreInViewerInPixels[ 0 ] = bdvHandle.getViewerPanel().getDisplay().getWidth() / 2.0;
+		windowCentreInViewerInPixels[ 1 ] = bdvHandle.getViewerPanel().getDisplay().getHeight() / 2.0;
+
+		adaptedViewerTransform.translate( windowCentreInViewerInPixels );
+
+		return adaptedViewerTransform;
+	}
+
+	public AffineTransform3D getAdaptedCenterTransform()
```
NicoKiaru, line 60: I would be in favor of adding a Run function like the other actions. The run function would center the BdvHandle to the given coordinates. Also it would be nice if you can give an Array of BdvHandle in order to recenter a set of BdvHandle.

The AffineTransform could be returned with a getter method

src/main/java/sc/fiji/bdvpg/bdv/navigate/ViewerTransformChanger.java
```diff
@@ -38,31 +39,53 @@
  *
  * See ViewTransformSetAndLogDemo for a usage example
```
NicoKiaru: Explain in the header why there's an animation duration, and how setting a negative (or 0?) value leads to instant displacement.

src/main/java/sc/fiji/bdvpg/scijava/command/bdv/ViewTransformatorCommand.java
```diff
@@ -78,6 +78,6 @@ public void run() {
         affineTransform3D.rotate(1, rotateAroundY);
         affineTransform3D.rotate(2, rotateAroundZ);
 
-        new ViewTransformator(bdvh, affineTransform3D).run();
+        new ViewerTransformChanger(bdvh, affineTransform3D, true, 0).run();
```
NicoKiaru, line 81: Do you think it can be useful to have another constructor wiht true and 0 being the default values ?

NicoKiaru (commented):
HI @tischi , I've done a few comments. Let me know what you think.

src/main/java/sc/fiji/bdvpg/scijava/command/bdv/ViewTransformatorCommand.java
tischi: Yes, I will do this.

src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
tischi: OK, I see what you mean. 

But then this method also needs the `animationDurationTime` input parameter of the `ViewerTransformChanger`.
Feels a bit like not so good separation of concerns to me.
In this sense keeping methods that _create_ viewer transforms (there could be many) and that _apply_ the viewer transform (maybe we have only one, but there could be also more in the future) separate feels better to me. What do you think?

I think the only reason one would need just one class that does both would be for the UI, and there I think we anyway would need to create a `Command` that then would wrap two calls like this: `new TransformApplier( new TransformCreator.get() ).run()`, isn't it?

So, in other words, I would prefer to put directly executable actions into `Command`s and keep things more separate on the backend API side.

src/main/java/sc/fiji/bdvpg/bdv/navigate/ViewerTransformChanger.java
tischi: ok!

src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
tischi: Docu: OK
Runnable: I don't think this should be runnable (see my answer below).

tischi: ...but we could make it `Callable` if that helps, returning an `AffineTransform`.

src/main/java/sc/fiji/bdvpg/scijava/command/bdv/ViewTransformatorCommand.java
tischi: But I think we should not put `concatenateTransform = true` as a default, because I was in fact surprised when I saw that it was concatenating. I would thus put no default there but require it in the constructor.
Ok?

tischi:
@NicoKiaru 
I saw you reacted to one of my answers. Are you still thinking about the other comments or did you overlook them? No rush, just to know whether I should still wait... ;-)

NicoKiaru:
Yep doing this slowly in parallel of many other stuff.

src/main/java/sc/fiji/bdvpg/bdv/navigate/AdaptedViewCenterTransformCreator.java
NicoKiaru: Ok, then why not BdvHandleHelper (this class may already exists) which gives  a few useful static function (centerViewOn, rescale, etc.) and only one or few actions which apply the transform on one or several BdvHandles ?

tischi:
@NicoKiaru 
I tried to change according to our discussion.

NicoKiaru:
thanks!

PR #162, opened 2020-11-04
Add SpatialSourceAndConverterSubsetter

tischi:
Let me know what you think. 
For my use cases this is working and useful and I would thus like to add it to our playground.
We could write into the description for which cases this will not work, but I don't know enough about the warped and transformed source to be able to judge this myself.


NicoKiaru:
Just on a general level : I'm not against it, but I think that we should try to avoid making this repository much much bigger.

I feel like the goal would be to maintain it in good shape and easy to integrate in other repositories because it is accessible via scijava maven.

Here we originally wanted to have everything as actions as Runnable + demo + GUI in commands. That's already the second class which do not respect this rule.

Why not having a separate repo ( I already have bigdataviewer-biop-tools) for things which are not well settled or where I'm the only current user for some actions ?

I think the goal here is to maintain this repo with minimal core functions. So I'm a bit reluctant to include new classes in the wild.

tischi:
> I think the goal here is to maintain this repo with minimal core functions. So I'm a bit reluctant to include new classes in the wild.

I think we need this functionality (I would even say it is a core functionality), otherwise how do we know which sources to act on in the (right-click) context menu?

> Here we originally wanted to have everything as actions as Runnable + demo + GUI in commands. That's already the second class which do not respect this rule.

As mentioned, I think the Command's are naturally the Runnable's. But I don't understand why all classes inside this repo should be `Runnable`? If a class is subsetting sources I think it should mainly have a function that returns a `List< SourceAndConverter >`, isn' it? But I am happy to add a `run` method, I just don't intellectually understand why this is needed. 

For example here: https://github.com/bigdataviewer/bigdataviewer-playground/blob/1c6c09fc0fe460e4768edea9841b931c28d97364/src/main/java/sc/fiji/bdvpg/sourceandconverter/transform/SourceResampler.java#L60

At first glance, I don't see why we need the run method there.

NicoKiaru:
This is just for convention. I know that for some actions, the run method is useless. Still this is what we decided in the first place. I originally suggested that we use scijava Commands everywhere, but it was decided that we want to be able to use actions independently of scijava Commands. I agreed with that, and we decided to keep actions at the most granular level while using commands for combining actions which would make sense to be executed sequentially.

I agree that the Runnable interface is not necessary in all actions, but then I would tend towards two things:

- either add these functionalities into a complete static helper class (there's one for SourceAndConverter, one for BdvHandle, let's add some if needed)

- or keep the convention, even if it may look stupid. One advantage that I see with the Runnable interface, is that it can be used uniformly in these ClickBehaviour which can be added to BigDataViewer.

I'd rather not have something completely in between a static class and an 'action' : an action which finally has no run method, or something not included in a static helper class.

I tried to stick to the convention so far, and included as much documentation and, demo and tests as I could. I'm clearly more messy in other repos, but it would be nice if we manage to keep this repo consistent and also potentially easy to maintain so without too many classes, trying to keep it slim.

For this particular aspect of subsetter : I would prefer having an extra method in the SourceAndConverterHelper class rather than of a full subsetter class.

This method:
https://github.com/bigdataviewer/bigdataviewer-playground/blob/1c6c09fc0fe460e4768edea9841b931c28d97364/src/main/java/sc/fiji/bdvpg/sourceandconverter/SourceAndConverterUtils.java#L479

Could be renamed with `isSourceAndConverterPIxelBlack`

And then an extra method with the other way of finding if a source is present:

could be named with `isSourceAndConverterInRAI` and implements what you want.

So is then just an extra class, and then the subset can be easily done with it.

In order to have a consistent core repo (I'm clearly much more messy in other repos). But for this base repo, let's try to make as much effort as possible.

tischi:
@NicoKiaru 
I tried to change according to discussion, please have a look.

NicoKiaru:
Cool!

PR #202, opened 2021-04-12
Context menu

tischi:
WIP on making the context menu more configurable


tischi:
@NicoKiaru I don't get the following:

In `SourceAndConverterService` there is this method:

```
   void registerDefaultActions() {
        this.registerAction("Display names", (srcs) -> {
            for (SourceAndConverter src:srcs){
                System.out.println(src.getSpimSource().getName());
            }});

        context().getService(PluginService.class)
                .getPluginsOfType(BdvPlaygroundActionCommand.class)
                .forEach(ci ->
                    registerScijavaCommandInfo(commandService.getCommand(ci.getClassName()))
                );

    }
```

Now, a `BdvPlaygroundActionCommand` does _not_ necessarily implement `Consumer<SourceAndConverter[]>`, however `SourceAndConverterService.getAction( String name )` returns a `Consumer<SourceAndConverter[]>`. How does that work?

NicoKiaru:
> Now, a `BdvPlaygroundActionCommand` does _not_ necessarily implement `Consumer<SourceAndConverter[]>`, however `SourceAndConverterService.getAction( String name )` returns a `Consumer<SourceAndConverter[]>`. How does that work?

I think it does, but then if there's nothing to do with the sources, they are just ignored, like in the expression below:

```
(sources) -> println("Hello")
```

If you look at `registerScijavaCommandInfo`, you'll find this case:

https://github.com/bigdataviewer/bigdataviewer-playground/blob/38bcf4e30ddff1e8b5ef40c8ecf077a9fb54d0b3/src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterService.java#L441-L444

tischi:
OK, I see, you just wrap it into a lambda! Clever 😉 

Should we officially support this like this?

```
    public void registerAction(String actionName, final BdvPlaygroundActionCommand action) {
        registerAction( actionName, (sacs) -> { commandService.run( action.getClass(), true ) ;} );
    }
```

Would that work?

NicoKiaru:
Hmm, not sure finally. It would be preferable to re-use registerScijavaCommandInfo internally. With registerScijavaCommandInfo, the sources are automatically consumed if there's a single compatible input parameter.

NicoKiaru:
You can use this : https://github.com/bigdataviewer/bigdataviewer-playground/blob/38bcf4e30ddff1e8b5ef40c8ecf077a9fb54d0b3/src/main/java/sc/fiji/bdvpg/scijava/services/SourceAndConverterService.java#L461

NicoKiaru:
>     public void registerAction(String actionName, final BdvPlaygroundActionCommand action) {
>         registerAction( actionName, (sacs) -> { commandService.run( action.getClass(), true ) ;} );
>     }

Also here you allow to name it differently from in the actual name of the command, so it's not a good idea.

tischi:
OK! Thanks! Maybe we could put all the context menu stuff into a separate service, to make more clear which methods are related to the context menu?

NicoKiaru:
I wouldn't create another service, but rather refactor the way context menu are created, as you mentioned in the other issue.

tischi:
@NicoKiaru I added a `MinimalBdvCreator`. Could you check whether this would work for you? If yes, maybe we can use this one and you could register the `Behaviours` that you need outside of that class?

NicoKiaru (commented):
I'm really not a fan of this static name field.

tischi:
I changed the static name field back in the latest commit, even tough I am a real fan of it :-)

tischi:
@NicoKiaru This seems to work for me now in MoBIE, I don't think I would need other features for now. Could we merge it?

NicoKiaru:
There are a few conflicts, shouldn't be too problematic. Can you merge master and resolve them ?

tischi:
Now looks good I think.

NicoKiaru:
👍 Merge when you want

PR #215, opened 2021-04-28
Implement SourceAndConverterFromSpimDataCreator

tischi:
@NicoKiaru 

I refactored the code such that we are more flexible in creating SACs from SpimData.

This now also allows separating SAC creation from SAC registration.

Let me know what you think!


NicoKiaru (approved):
I didn't tested it all, but looks right!

NicoKiaru (approved):

Issue #201, opened 2021-03-28
Get rid of all capital letters in Commands!


NicoKiaru:
the macro recorder does not record the case of variable.

So : an input parameter named "numTimePoints" will be recorded "numtimepoints"

That's super annoying - we need to get rid of them, and maybe use a dash convention "numTimepoint" -> "num_timepoints", at least in scijava commands

tischi:
but the macro will still work even if the capitalization is wrong, doesn't it?

NicoKiaru:
ah, the macro probably (I didn't test it). I had issues in groovy.
Here's what I did which annoyed me:
* run the macro recorder
* perform an action with the right click
* use the output from the recorder to run a command with a groovy script : `commandService.run(MyCommand.class, true, "arg1", val1, "arg2", val2)` . This doesn't work because arg1 and arg2 from the recorder are not capitalized, with the command service expects the right capitalization

tischi:
Yes, I know, I have an "issue" about this on the forum already, but I cannot find it right now.
I don't like underscore in Java, I would go for all lower case as a work-around, but I think you should complain in the scijava gitter channel about this, as this is bug, imho.

NicoKiaru:
👍 Let's go for all lower case. Can you link the post to the issue ?

PR #216, opened 2021-04-29
Put command variable parameter to lower case

NicoKiaru:
Ok with that @tischi ? It's what we've decided for the issue #201


tischi:
okidoki!
it's a shame we need to do this though.

Issue #180, opened 2021-03-23
Simplify API


tischi:
```
SourceAndConverterServices.getSourceAndConverterDisplayService()
```

Maybe rename to:

```
SourceAndConverterServices.getDisplayService()
```

NicoKiaru:
Yes, we could also think whether we want to mention that it's a bdv display service

So `getBdvDisplayService`

tischi:
> Yes, we could also think whether we want to mention that it's a bdv display service

Good idea, because there may also be a `getBvvDisplayService` or `getSceneryDisplayService` at some point 😄

Issue #199, opened 2021-03-27
Blending of sources / custom converter


NicoKiaru:
Here we can discuss plans for custom Converters and Projectors for bdv.

I just drop a few questions to start with:
* how to handle transparency ?
** at the converter level ? (identical transparency for multiple bdvs)
** at the projector level ? (transparency is a property of the current bdv)
* do we want layers ? can they be reordered ? what properties should layers have (enabled / disabled / transparency)
* what about overlays ?
* are layers sorted ?
* how is the projector aware of layers ?
* how to serialize converters (potentially layers ?)
* how to compute correctly blending ?

We should keep an eye on:
* bdv performance (we need to measure - we need benchmarks)
* making sure the events are well handled (if we go for layers, make sure the source are removed from layers if they are removed from the bdv window) (we need tests)

ping @tischi @tpietzsch @frauzufall @constantinpape @martinschorb

tischi:
References:
- https://en.wikipedia.org/wiki/Alpha_compositing#cite_note-3
- https://github.com/saalfeldlab/paintera/tree/master/src/main/java/org/janelia/saalfeldlab/paintera/composition

Issue #196, opened 2021-03-26
ProjectionMode


tischi:
@NicoKiaru 

Are you using this in your projects?

```
public class Projection
{
	public static final String PROJECTION_MODE = "Projection Mode"; // underscore necessary for valid xml element to store in @see DisplaySettings
	public static final String PROJECTION_MODE_SUM = "Sum";
	public static final String PROJECTION_MODE_AVG = "Average";
	public static final String PROJECTION_MODE_OCCLUDING = "Occluding";

	public static final String PROJECTOR = "Projector";
	public static final String MIXED_PROJECTOR = "Mixed Projector";
	public static final String SUM_PROJECTOR = "Sum Projector";
	public static final String AVERAGE_PROJECTOR = "Average Projector";
}
```

I would like to clean this up by making it an `enum` for the projection mode.

NicoKiaru:
Please do - from what I remember we had some troubles with scijava commands fields

tischi:
> from what I remember we had some troubles with scijava commands fields

damn, yes....

NicoKiaru:
Related : PR https://github.com/bigdataviewer/bigdataviewer-playground/pull/198

NicoKiaru:
I think this is now obsolete because the projector options are not dealt with anymore in this repository, but in bigdataviewer-playground-display

Issue #140, opened 2020-07-13
 edit mode and group mode kill each other


martinschorb:
Hi,

The BDV window gets locked up and reacts only very delayed when I change between edit mode and group view.

NicoKiaru:
thanks for noting this.

The group mode is kind of messed up. I hope that https://github.com/bigdataviewer/bigdataviewer-core/pull/99 solves it. I have to try

The edit mode also had an independent issue which I adressed yesterday but did not push to the update site yet.

NicoKiaru:
Many it depends on the number of sources displayed - if you have many, the overlay can take a long time to be drawn. Regarding group mode : if the overlay is not in any group, just like any other source, it's not displayed.

If you still have this issue, please ping me @martinschorb , otherwise I'll probably close the issue

NicoKiaru:
Not sure this still happens, closing for now because I have no way to reproduce this easily

PR #219, opened 2021-05-06
Refactor display options

NicoKiaru:
@tischi, 

Here's the big refactoring. It removes a lot of stuff with bdv playground, but together with https://github.com/NicoKiaru/bigdataviewer-spimdata-extras and https://github.com/NicoKiaru/bigdataviewer_playground_display we get a better separation of tasks, the same functionalities as before, and an easier and more flexible to extend them.

This will require a bit of time to adapt your code, but it'll be worth.

We should be able to solve, or provide an easier way to fix:
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/199 -> done in bigdataviewer_playground_display
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/196 ->  done in bigdataviewer_playground_display with more flexibility
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/140
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/136
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/131
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/117
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/95
* https://github.com/bigdataviewer/bigdataviewer-playground/issues/94

The two things which are allowing this is the serialization of the suppliers of Bdv, which allows to store preferences conveniently for how the bdv is build. I also added plugins which can basically handle any sourceandconverter operation based on a tag of xml of spimdata objects, when spimdata objects are loaded or saved ( think displaysettings, projection, and also custom converters for label images)

Maybe we should tackle https://github.com/bigdataviewer/bigdataviewer-playground/issues/180 while we're at it.


NicoKiaru:
Ok @tischi , let's go for it!

NicoKiaru:
Hmm I'll fix #180 before

PR #220, opened 2021-05-12
Add CroppedSource

tischi:
Hi @NicoKiaru,
Here's my work in progress (not ready to be merged yet, I did not test it).

### DIFF
--- a/src/main/java/bdv/util/CroppedSource.java
+++ b/src/main/java/bdv/util/CroppedSource.java
@@ -35,6 +35,7 @@ import bdv.viewer.Interpolation;
 import bdv.viewer.Source;
 import mpicbg.spim.data.sequence.VoxelDimensions;
 import net.imglib2.FinalInterval;
+import net.imglib2.FinalRealInterval;
 import net.imglib2.Interval;
 import net.imglib2.RandomAccessible;
 import net.imglib2.RandomAccessibleInterval;
@@ -61,18 +62,21 @@ public class CroppedSource< T extends NumericType<T> & NativeType<T>> implements
     private final Source<T> wrappedSource;
     private final String name;
     private final RealInterval crop;
+    private final boolean cropMinToZero;
     protected final DefaultInterpolators< T > interpolators;
     private HashMap< Integer, Interval > levelToVoxelInterval;
 
-    public CroppedSource( Source<T> source, String name, RealInterval crop )
+    public CroppedSource( Source<T> source, String name, RealInterval crop, boolean cropMinToZero )
     {
         this.wrappedSource = source;
         this.name = name;
         this.crop = crop;
+        this.cropMinToZero = cropMinToZero;
         this.interpolators = new DefaultInterpolators();
 
         initCropIntervals( source, crop );
     }
+
 
     private void initCropIntervals( Source< T > source, RealInterval crop )
     {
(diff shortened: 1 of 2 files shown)


tischi:
@NicoKiaru 
I have added a working Demo. What do you think?
I would add a `SourceCropper` to make this work on a `SourceAndConverter`.

NicoKiaru:
We can have probably a static function which does the same using empty source and resampler. The problem with CroppedSource, is that it's not serializable currently, and I want to have all kinds of sources serializable in the playground.

Both EmptySource and ResampledSource are serialiazable, while CroppedSource, which as far as I understand is one of them with less flexibility, is not.

We could make it serializable, but why would we want to do that ?

tischi:
Why is CroppedSource not serializable?

NicoKiaru:
Did you try it ? There are some serialization tests in the state saver and loader. You can add your cropped source demo and see if it works. Maybe it is working.

tischi:
I think it should be working because I could make it such that the only fields of the CroppedSource class are:
```
Source wrappedSource;
double[] min;
double[] max;
```
The `wrappedSource` can be serialized by `SpimSourceAdapter` and the `double[]` can anyway be serialized.

Right now however it does not work, because in the `SourceAndConverterAdapter` code there is this line:

```
if (!sourceSerializers.containsKey(sourceAndConverter.getSpimSource().getClass())) {
            System.out.println("Unsupported serialisation of "+sourceAndConverter.getSpimSource().getClass());
            throw new UnsupportedOperationException();
        }
```

And of course here is no specific adapter for `CroppedSource`.

However, as pointed out above, I don't think we need a specific adapter for `CroppedSource`, because this can be serialized as is. 

Can one maybe change the code in `SourceAndConverterAdapter` to simply `try` to serialize a given `Source` and then throw an error if that does not work?

I think that would be very cool in order not to have to implement a custom adapter for classes that can be serialized as is.

NicoKiaru:
Indeed, I think there's a few improvements which can be made on the serialization now that I understand a bit better how it all works...

But "circular" dependencies still needs to be handled in a specific way.

I still do not understand why you need this cropped source when the resampled source and the empty source can do the same job and already has:
* a working serializer
* [an action](https://github.com/bigdataviewer/bigdataviewer-playground/blob/master/src/main/java/sc/fiji/bdvpg/sourceandconverter/transform/SourceResampler.java)
* [is properly inspected](https://github.com/bigdataviewer/bigdataviewer-playground/blob/master/src/main/java/sc/fiji/bdvpg/scijava/services/ui/SourceAndConverterInspector.java)
* [can be searched for in the root source](https://github.com/bigdataviewer/bigdataviewer-playground/blob/18c32ee59d1d5305d98d24c232ce614732dada9b/src/main/java/sc/fiji/bdvpg/sourceandconverter/SourceAndConverterHelper.java#L742)
* and also in the experimental alpha projector support in bigdataviewer-playground-display

There are also command in biop tools which are using them.

For a full support, you will need to modify the inspector, add new commands and actions, modify the way we find the 'root' source recursively.

So I fear this will create more complexity, but if you think it's worth it, please merge. Also, you mentioned you were going to try this in MoBie before we would put it in the playground. Was there an issue when trying inside MoBiE ?

tischi:
> Also, you mentioned you were going to try this in MoBie before we would put it in the playground. Was there an issue when trying inside MoBIE ?

There was no issue, I just thought it's something that would be nice to have in the playground as it falls into the category "Simple reusable class that does something useful with a SourceAndConverter", something were I thought the playground would be a nice place to have it (even if we do not expose it in the Command's UI). And I was not aware of the cool serialization framework that you implemented and that newly added code needs to play well with this.

I also must admit I did not use the `ResampledSource` and `EmptySource` because I did not understand how to use it for only the cropping and it was thus easier to implement a `CroppedSource` (and as said I was not aware about the serialization framework). I think I would need your help here 😃 

Otherwise I can also simply put the `CroppedSource` into MoBIE. Maybe this is better because for MoBIE I need to (de)serialize into `double[] min, max` for the crop which is easier if the fields exist in the `CroppedSource` class. Long term would be of course very cool to merge the serialization done in MoBIE and the playground but that's for the next hackathon 😉.

NicoKiaru:
Oki doc, I agree this Empty Source is not straightforward. I'll try to see if there's a convenient way to use it, or if cropped source is the way to go.

Cheers

PR #240, opened 2021-08-18
Add a ZSlider in bdv based on ray casting

NicoKiaru:
This PR adds an actions which adds a ZSlider in a Bdv window - and a demo of it (RayCastDemo) 

Because the orientation in Bdv can be anything, the way the slider works is the following:
a 'virtual' ray is cast at the center of the bdv window and each plane which intersects this ray corresponds to a position within the slider.

All positions from all sources are finally sorted along Z.

The slider works 'directionally' : if the user moves the bdv window, the slider is updated to its new z position, if the user moves the slider, the bdv window is shifted in z according to the new position.

The slider works only for 'standard' sources ( = 3D Rai with affine transforms ), and not for warped sources.

### DIFF
--- a/src/main/java/sc/fiji/bdvpg/bdv/BdvHandleHelper.java
+++ b/src/main/java/sc/fiji/bdvpg/bdv/BdvHandleHelper.java
@@ -29,10 +29,14 @@
 package sc.fiji.bdvpg.bdv;
 
 import bdv.tools.brightness.ConverterSetup;
+import bdv.util.BdvFunctions;
 import bdv.util.BdvHandle;
+import bdv.util.BdvOptions;
+import bdv.util.BdvOverlay;
 import bdv.viewer.Source;
 import net.imglib2.*;
 import net.imglib2.realtransform.AffineTransform3D;
+import net.imglib2.type.numeric.ARGBType;
 import net.imglib2.util.Intervals;
 import net.imglib2.util.LinAlgHelpers;
 import org.scijava.cache.CacheService;
@@ -46,6 +50,7 @@ import sc.fiji.bdvpg.bdv.config.BdvSettingsGUISetter;
 import sc.fiji.bdvpg.scijava.services.SourceAndConverterBdvDisplayService;
 
 import javax.swing.*;
+import java.awt.*;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
 import java.io.File;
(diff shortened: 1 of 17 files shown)


NicoKiaru:
For tischi : only the last commit really matter

tischi:
@NicoKiaru I tried it out but when moving the z-slider, nothing happens for me.

image

NicoKiaru:
Hmm. The ray is cast at the center of the screen. If there's no image crossing it, then no movement happens. From the image it's unclear whether the cross actually 'touches' an image. But this could be something else...

tischi:
Cool! Moving a source onto the cross works!

What's the extend of the range supposed to be?

I did Shift+X on a source and scrolled and it felt like I could go a bit too far, i.e. into neighboring sources.

NicoKiaru:
The range is 'infinite'. A ray is cast at the center of the window, and all planes from any source intersecting this ray will have a 'plane index'. All plane index are sorted and put into the slider. This means that all spacings are not equal (unless the ray crosses only one source). Sometimes you will 'jump' from the end of one source to the beginning of the next one.

tischi:
What do you mean by infinite?

image

I am at the end of the slider but still within our universe, even within the sources.

NicoKiaru:
If there's no source further away, then you stop at the last intersecting plane. But it's 'potentaially infinite' in the sense that if there is an image further away, the slider would have extra positions. It's not a fixed range with a fixed step size. 

Here's a small drawing showing what the positions would look like in the case of 3 randomly placed sources:
image

Even if the positions in the slider are evenly spaced, these positions are not evenly spaced in the real physical coordinates. Also if there are intertwined planes, they are sorted along their coordinates along the z axis.

Is it clearer ?

tischi:
Thanks! Yes, this is awesome, dude!

The only question I would have whether (a) it is only the central ray that is used to determine when a source ends or whether (b) it is the whole viewer window? 

image

I think already (a) is awesome!

NicoKiaru:
So far it's the central ray. There are a few things which could be tried - like casting rays on the 4 corners of the image and try to 'merge' them together, but I'm sure how to do that.

What we could also do is to add a button that rotates the view in order to make it perpendicular to the current nearest plane. Then maybe the slider feeling would be more natural.

NicoKiaru:
It should not be too difficult to put all planes of a RAI as long as the ray intersects at least one plane.

tischi:
I would merge it as is and then make issues about your suggestions.
Should we merge?

PR #255, opened 2022-06-19
Simplify creation of converter setups

tischi:


tischi:
@NicoKiaru 
For me the changes work (and in fact I need them for MoBIE, where we have quite unusual Converters...).

NicoKiaru:
Hi tischi,

Three small things and you can merge:

* the last commit is extremely weird : are you sure you can't find the reason ?
* since it's private methods, no need to keep them with deprecated annotation. The @deprecated annotation only makes sense for public methods.
* have you run the tests ? Do they all pass ?

tischi:
> the last commit is extremely weird : are you sure you can't find the reason ?

I changed it back and it also works (was something weird with IntelliJ...)


>    since it's private methods, no need to keep them with deprecated annotation. The @deprecated annotation only makes sense for public methods.

Good point! I did not realise they were private. I removed them.

>    have you run the tests ? Do they all pass ?

Yes.

PR #269, opened 2024-02-06
BREAKING: bump to bigdataviewer-vistools-1.0.0-beta-34

axtimwalde:
* remove VolatileSource in favor of the upstream VolatileSource


NicoKiaru:
I'm just wondering: in the VolatileSource from this repo (the old one), I made sure that `VolatileViews.wrapAsVolatile` was called only once.

With the new version, `VolatileViews.wrapAsVolatile` is called many many times. Could this be an issue ?

PR #280, opened 2026-02-19
Big refactor

NicoKiaru:
Hi @tischi,

I plan to do this gigantic refactoring. All the functionalities are still there, but I've simplified the names and standardized them. Essentially all these 'SourceAndConverter' are renamed to Source as much as possible.

On top of that I've refactored some other names and move packages here and there to be more consistent.

I'm happy to check and do a PR of what needs to be changed in mobie.

If you have time and have suggestions about the big refactoring I take them! But if you don't, them I guess you'll have to trust me in the fact that it is an improvement ;-).

I think, besides me, you're the only one that uses directly this repo as a dependency in other projects.

(Note that bigdataviewer-image-loaders is independent and unchanged by this refactoring, and I don't plan to change it because it's reasonably clean in terms of structure).

Cheers


tischi:
Hi, I trust your refactoring and I would be grateful to have a PR against MoBIE.

